PR #4, opened 2021-10-14
Pia 1854 release GitHub actions

a-szotyori:
Adds release workflows to release to our own maven repository.

Once we get publisher access to Maven Central I'll update the release workflows to publish our releases there.


.github/workflows/health-api-library.check.yml
```diff
@@ -1,8 +1,13 @@
 name: Check Health API Library
+
 on:
   push:
     paths:
       - 'health-api-library/**'
+    tags-ignore:
+      - '**'
```
zladzeyka, line 8: 🙈what tags are ignored , all?

a-szotyori: Exactly. The check workflow should not run for tags because we'll start it in the release workflows and only do the release if the check workflow is successful.

PR #21, opened 2021-12-06
refactor(bank-sdk): Update example app package names

a-szotyori:
The test payment provider was updated to use the new package name
so we could finally replace the old `net.gini.pay.appscreenapi` and
`net.gini.pay.appcomponentapi` package names with the new ones:
`net.gini.android.bank.sdk.screenapiexample` and
`net.gini.android.bank.sdk.componentapiexample`.


zladzeyka (approved):
🎉

PR #32, opened 2021-12-22
fix(health-sdk): Preserve preview zoom position when showing/hiding the keyboard

a-szotyori:


zladzeyka (changes requested):
Screenshot 2022-01-05 at 15 00 53

I didn't manage to fix this issue and run the app.

PR #40, opened 2022-01-19
refactor(bank-sdk): Show how to use suspending functions from Java

a-szotyori:
Added a Java implementation of the `PayViewModel` called
`PayViewModelJava` (kotlin version was renamed to
`PayViewModelKotlin`) which shows how to use the suspending
functions from Java.


zladzeyka (commented):
👏🏻

PR #42, opened 2022-01-27
Pia 2023 switch to health api in library

a-szotyori:
Sorry for this huge PR... I had to touch a lot of files to allow the core api library to be extended by the health and bank api libraries. This way I could move health api only calls to the health api library and the same for the bank api library.

To not make this PR even bigger I didn't touch the health and bank sdks so their tests will fail. After the review I can either continue pushing into this PR or we merge this and I continue in a new branch and create a new PR.


zladzeyka (approved):
😎💻👏🏻

PR #55, opened 2022-04-29
Pia 2460 return assistant customization guide

a-szotyori:


zladzeyka (approved):

bank-sdk/sdk/src/main/res/values/styles.xml
```diff
@@ -383,6 +394,10 @@
 
     <style name="GiniCaptureTheme.DigitalInvoice.LineItemDetails.Total.TextStyle" parent="Root.GiniCaptureTheme.DigitalInvoice.LineItemDetails.Total.TextStyle" />
 
+    <style name="GiniCaptureTheme.DigitalInvoice.LineItemDetails.Total.Info.TextStyle" parent="Root.GiniCaptureTheme.DigitalInvoice.LineItemDetails.Total.Info.TextStyle" />
```
zladzeyka, line 397: @a-szotyori I've noticed in the docs  `GiniCaptureTheme...` used, sometimes  `Root.GiniCaptureTheme...` is it the same?

zladzeyka (approved):

bank-sdk/sdk/src/main/res/values/styles.xml
a-szotyori: Same answer as in Slack (sorry for not seeing your question here first 🙈 ):

This is how style inheritance is done. Using the `Root` style as the parent allows clients to customise the style while also inheriting all our default style settings. Without having the `Root` style as parent clients would end up deleting all our default style settings when they customise a style.

(I learned this from the Google android developers through their Material Design library, [they use the prefix ](https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/textview/res/values/styles.xml#L19)`Base` instead of `Root`).

PR #60, opened 2022-06-10
Pia 2290 onboarding

a-szotyori:
This is a huge PR. I recommend checking it out locally and trying out the onboarding with the screen api example app.

Besides the onboarding UI update it contains the top and bottom navigation bar UI injection and the custom onboarding illustration injection.

Code for the UI injection is in the packages: `net.gini.android.capture.view` and `net.gini.android.capture.onboarding.view`.

Code for the onboarding is in the `net.gini.android.onboarding` package.

The `GiniCapture` and `GiniCapture.Builder` were also update to allow UI injection and enabling/disabling the bottom navigation bar.

Customising the UI colors and fonts was simplified by using a [color system with named colors](https://github.com/gini/gini-mobile-android/blob/PIA-2290-onboarding/capture-sdk/sdk/src/main/res/values/colors.xml#L119) and a [typography system based on Material UI](https://github.com/gini/gini-mobile-android/blob/PIA-2290-onboarding/capture-sdk/sdk/src/main/res/values/typography.xml). If clients override any of these, then the UI will be updated on all screens and clients don't have to customise each screen separately.


zladzeyka (approved):
@a-szotyori That's amazing work! 👏🏻🎉

PR #63, opened 2022-07-01
Pia 2652 customisation guide

a-szotyori:
We will embed figma projects instead of adding screenshots and writing all the details into the page. 

We will have all the screenshots and details in figma because it's easier to maintain.

The figma project for the Android Capture SDK is here: https://www.figma.com/file/vrXbRphncnduEHM5eoFLDm/Android-Gini-Capture-SDK-2.0.0-UI-Customisation?node-id=0%3A1

Please note that it is versioned and whenever we have changes we need to duplicate it and use the Capture SDK's version in its title. We then embed that version into the page.


zladzeyka (approved):
Great success 👏🏻👍🏻

PR #66, opened 2022-07-07
chore: Clarify conventional commit type descriptions

a-szotyori:
As our team grows I think it's helpful to clarify when to use which commit type.

Any improvement feedback is very welcome. 😃 

Once we agree on the final version, I'd update it in the iOS monorepo as well.


zladzeyka (approved):
👍🏻

PR #67, opened 2022-07-12
Pia 2753 pay button

a-szotyori:
* Enable pay button when total > 0.
* Change pay button title when none of the items are selected.


zladzeyka (changes requested):
Looks 😊 good and tests 👏🏻🎉
The only edge case with other costs 0 left. I've attached iOS screens.
We need to disable pay button when total is 0
IMG_007AE6930962-1

IMG_1901B868D3DD-1

PR #68, opened 2022-07-13
fix(bank-sdk): Don't handle input in `CaptureFlowActivity` on activity restart

a-szotyori:
Handling the input on activity restarts caused the camera activity to be launched again when the
digital invoice activity returned with a result.

PIA-2785


zladzeyka (approved):
👍🏻

PR #69, opened 2022-07-15
feat(bank-sdk): Line item quantity input is limited to be between 1 and 99,999

a-szotyori:
PIA-2778

### DIFF
--- a/bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/capture/digitalinvoice/details/LineItemDetailsFragment.kt
+++ b/bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/capture/digitalinvoice/details/LineItemDetailsFragment.kt
@@ -13,8 +13,6 @@ import androidx.core.content.res.ResourcesCompat
 import androidx.core.view.isVisible
 import androidx.fragment.app.Fragment
 import com.google.android.material.textfield.TextInputLayout
-import net.gini.android.capture.internal.util.ActivityHelper.forcePortraitOrientationOnPhones
-import net.gini.android.capture.network.model.GiniCaptureReturnReason
 import net.gini.android.bank.sdk.R
 import net.gini.android.bank.sdk.capture.digitalinvoice.LineItem
 import net.gini.android.bank.sdk.capture.digitalinvoice.ReturnReasonDialog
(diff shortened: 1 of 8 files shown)


bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/capture/digitalinvoice/details/LineItemDetailsFragment.kt
```diff
@@ -225,9 +245,20 @@ class LineItemDetailsFragment : Fragment(), LineItemDetailsScreenContract.View,
...
         setInputHandlers()
     }
 
+    private fun addGlobalFocusChangeListener() {
+        // Workaround for the bug in TextInputEditText where the focus change listener is not called
+        // when using end icons: https://github.com/material-components/material-components-android/issues/1015
+        binding.root.viewTreeObserver.addOnGlobalFocusChangeListener { oldFocus, _ ->
+            if (oldFocus == binding.quantity) {
+                presenter?.onQuantityInputFieldFocusLoss(binding.quantity.text.toString())
+            }
```
zladzeyka, line 258: It's still a bit not clear what is it the focus change here. Please see my video attached. It's related to the focus change 

https://user-images.githubusercontent.com/3911816/179530868-7473ea70-6205-4f79-bfb9-5028aadbfb07.mp4

zladzeyka (changes requested):
🙏 Please, find my video about the focus change video attached.

bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/capture/digitalinvoice/details/LineItemDetailsFragment.kt
a-szotyori: I can make the font size decrease to a minimum. Won't guarantee that it will always fit, but for the majority of cases it should be ok.

zladzeyka: > I can make the font size decrease to a minimum. Won't guarantee that it will always fit, but for the majority of cases, it should be ok.

Could be great 👍

a-szotyori:
I made the font size of the quantity and the total price labels dynamic.

zladzeyka (approved):
👏🏻🎉😊

PR #84, opened 2022-08-24
Pia 2656 support predictive back gesture

a-szotyori:
Note: please review only the commits [ed57c9f](https://github.com/gini/gini-mobile-android/pull/84/commits/ed57c9f1481c136fced8b739eae6cc707734c0d8) and newer.

I replaced all `onBackPressed()` overrides with AndroidX's `OnBackPressedCallbacks`.

The changes have no visible effect. Every back button press or back swipe gesture should work as usual even on Android 13.

The predictive back gesture feature is disabled in Android 13 and can be enabled for testing purposes in the developer tools only. 

It's been a while since android recommends using `OnBackPressedCallbacks` so even though these changes won't have any effect now we are reducing future technical debt.

https://developer.android.com/guide/navigation/predictive-back-gesture
https://android-developers.googleblog.com/2022/07/prepare-your-app-to-support-predictive-back-gestures.html

### DIFF
--- a/bank-sdk/component-api-example-app/src/main/AndroidManifest.xml
+++ b/bank-sdk/component-api-example-app/src/main/AndroidManifest.xml
@@ -4,6 +4,7 @@
     package="net.gini.android.bank.sdk.componentapiexample">
 
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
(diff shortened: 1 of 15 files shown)


zladzeyka (approved):
👍🏻

PR #85, opened 2022-08-30
Pia 2482 fix camera resolution requirement

a-szotyori:
Unfortunately CameraX does not report all resolutions on some devices (Galaxy A41 or A5). I could reproduce the issue with this official [CameraX example app](https://github.com/android/camera-samples/tree/master/CameraXBasic), too.

Due to this our camera resolution requirements check fails because the largest picture size is only 3.7 MP and we require at least 8 MP.

Using the old camera api all resolutions are reported and our checks succeed.

The implemented solution is to try using CameraX and fall back to the old camera api, if it cannot be used. This fallback is implemented in the requirements checker and on the camera screen.


zladzeyka (approved):
Looks good to me!

PR #88, opened 2022-09-07
fix(bank-sdk): Launch single page review screen for "open with" if multi-page is disabled

a-szotyori:
Testing instructions: https://ginis.atlassian.net/browse/PIA-3095?focusedCommentId=19165


zladzeyka (approved):
👍🏻

PR #101, opened 2022-10-18
Pia 2982 analysis UI inject top bar

llevente:


a-szotyori (approved):
Nice! I see it is branched off of PR #100. Please don't merge it yet. We need to go through a small QA by our PO before merging.

PR #103, opened 2022-10-20
Pia 3145 image cropping

AdnanYupi:


capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/camera/photo/PhotoEdit.java
```diff
@@ -73,6 +85,39 @@ private void removeCompressionModifier() {
...
+    public PhotoEdit crop(Activity context, Rect aRect) {
+
+        byte[] originalBytes = mPhoto.getData();
+
+        try {
+            Bitmap originalBitmap = BitmapFactory.decodeByteArray(originalBytes,
+                    0, mPhoto.getData().length);
+
+            DisplayMetrics metrics = new DisplayMetrics();
+            context.getWindowManager().getDefaultDisplay().getMetrics(metrics);
```
a-szotyori, line 97: [Suggestion] The display size does not account for the navigation bar. Can you use the `gc_camera_preview_container` to get the size that accounts for the navigation bar?

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/camera/photo/PhotoEdit.java
```diff
@@ -73,6 +85,39 @@ private void removeCompressionModifier() {
...
+
+            int x1 = originalBitmap.getWidth() * aRect.left / metrics.widthPixels;
```
a-szotyori, line 99: [Request] The bitmap can be rotated and we need to account for that. The `mPhoto.getRotationForDisplay()` can be used to find out how the bitmap needs to be rotated to be in the right orientation for the cropping to work.

Rotation is always by 90 degree steps so this switch will cover all cases:
```
            switch (mPhoto.getRotationForDisplay() % 360) {
                case 0:
                    // bitmap and aRect have same rotation
                    break;
                case 90:
                    // bitmap is rotated 90 CCW compared to aRect
                    break;
                case 180:
                    // bitmap is upside-down compared to aRect
                    break;
                case 270:
                    // bitmap is rotated 270 CCW compared to aRect
                    break;
                default:
                    // this should never happen
                    break;
            }
```

For each case the crop values x1, y1, width1 and height1 need to account for the rotation.

If we don't then for example on the Galaxy S21 the cropping is wrong because the `getRotationForDisplay()` is `90` degrees:
Screenshot_20221021_182651
Screenshot_20221021_182710

a-szotyori (changes requested):
The direction is good. Some further adjustments are needed.

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/camera/photo/PhotoEdit.java
```diff
@@ -73,6 +85,39 @@ private void removeCompressionModifier() {
         }
     }
 
+    public PhotoEdit crop(Activity context, Rect aRect) {
+
+        byte[] originalBytes = mPhoto.getData();
```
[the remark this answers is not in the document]
AdnanYupi: @a-szotyori for some reason I thought that already all photo code is running on the background thread :D

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/camera/photo/PhotoCropModifier.java
```diff
@@ -0,0 +1,91 @@
...
+        this.aRect = aRect;
+    }
+
+    @Override
+    public void modify() {
+
+        if (mPhoto.getData() == null)
+            return;
+
+        synchronized (aRect) {
```
a-szotyori, line 29: [Request] Please use mPhoto as the monitor object. The other modifiers use it as well.

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/camera/photo/PhotoCropModifier.java
```diff
@@ -0,0 +1,91 @@
...
+                        break;
+                    case 180:
+                        rotation = 180;
+                        break;
+                    case 270:
+                        rotation = 270;
+                        break;
+                    default:
+                        break;
+                }
```
a-szotyori, line 46: [Suggestion] Can this be simplified to `float rotation = mPhoto.getRotationForDisplay() % 360`?

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/camera/photo/PhotoCropModifier.java
```diff
@@ -0,0 +1,91 @@
...
+
+                Matrix matrix = new Matrix();
+                matrix.postRotate(rotation);
+
+                Bitmap originalBitmap = BitmapFactory.decodeByteArray(mPhoto.getData(),
+                        0, mPhoto.getData().length);
+
+                Bitmap rotatedBitmap = Bitmap.createBitmap(originalBitmap, 0, 0, originalBitmap.getWidth(),
```
a-szotyori, line 54: [Comment] Nice straightforward solution! 👍

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/camera/photo/PhotoCropModifier.java
```diff
@@ -0,0 +1,91 @@
...
+                        originalBitmap.getHeight(), matrix, false);
+
+
+                int x1 = rotatedBitmap.getWidth() * aRect.left / mScreenSize[0];
+                int y1 = rotatedBitmap.getHeight() * aRect.top / mScreenSize[1];
+                int width1 = rotatedBitmap.getWidth() * aRect.width() / mScreenSize[0];
+                int height1 = rotatedBitmap.getHeight() * aRect.height() / mScreenSize[1];
+
+                matrix.postRotate(finalRotation(rotation));
```
a-szotyori, line 63: [Suggestion] If I'm not mistaken then this line reverses the rotation so that the cropped image has the same rotation as the original.

I did a bit of research and found that we can use `matrix.invert(matrix)` to reverse the rotation.

With `finalRotation()` the orientation was wrong when I tested a 270 rotation on a Galaxy S21 by changing the line 45 in `ActivitHelper` to `activity.setRequestedOrientation(ActivityInfo.SCREEN_ORIENTATION_REVERSE_PORTRAIT)`.

With `matrix.inverse(matrix)` it worked for 270 and other rotations as well (I uncommented the lines in `ActivityHelper.lockToPortraitOrientation()` to be able to test landscape orientations on the phone).

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/camera/photo/PhotoCropModifier.java
```diff
@@ -0,0 +1,91 @@
...
+
+                Bitmap cropped = Bitmap.createBitmap(rotatedBitmap, x1, y1,
+                        width1, height1, matrix, false);
+
+                ByteArrayOutputStream stream = new ByteArrayOutputStream();
+                cropped.compress(Bitmap.CompressFormat.PNG, 100, stream);
```
a-szotyori, line 69: [Request] We should use `JPEG` compression format. We don't have transparency and we need jpegs for exif data.

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/camera/photo/PhotoCropModifier.java
```diff
@@ -0,0 +1,91 @@
...
+
+                byte[] byteArray = stream.toByteArray();
+                mPhoto.setData(byteArray);
+
+                originalBitmap.recycle();
```
a-szotyori, line 74: [Clarification] Shouldn't `rotatedBitmap` be also recycled?

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/camera/photo/PhotoCropModifier.java
```diff
@@ -0,0 +1,91 @@
...
+
+                Bitmap cropped = Bitmap.createBitmap(rotatedBitmap, x1, y1,
+                        width1, height1, matrix, false);
+
+                ByteArrayOutputStream stream = new ByteArrayOutputStream();
+                cropped.compress(Bitmap.CompressFormat.PNG, 100, stream);
+
+                byte[] byteArray = stream.toByteArray();
+                mPhoto.setData(byteArray);
+
```
a-szotyori, line 73: [Request] Please also call `mPhoto.updateBitmapPreview()` and `mPhoto.updateExif()`. The first one is obvious and the second one is why we need jpegs. We keep a subset of exif data and store some custom meta information like sdk version in the exif.

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/camera/photo/PhotoEdit.java
```diff
@@ -73,6 +85,17 @@ private void removeCompressionModifier() {
         }
     }
 
+    public PhotoEdit crop(View mFrame, Rect aRect) {
+
+        int[] frameSize = {mFrame.getWidth(), mFrame.getHeight()};
+
+        PhotoCropModifier cropModifier = new PhotoCropModifier(mPhoto, frameSize, aRect);
```
a-szotyori, line 92: [Comment] Perfect, thank you! 👍

a-szotyori (changes requested):
Almost there 🙈 !

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/camera/photo/PhotoCropModifier.java
AdnanYupi: Yeah true. I used to set `finalRotation()` but as the matrix has inverse makes this obsolete.

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/camera/photo/PhotoCropModifier.java
AdnanYupi: Yep correct, it should

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/camera/photo/PhotoCropModifier.java
AdnanYupi: Awesome thanks.

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/camera/photo/PhotoEdit.java
```diff
@@ -89,7 +89,7 @@ public PhotoEdit crop(View mFrame, Rect aRect) {
 
         int[] frameSize = {mFrame.getWidth(), mFrame.getHeight()};
 
-        PhotoCropModifier cropModifier = new PhotoCropModifier(mPhoto, frameSize, aRect);
+        PhotoCropModifier cropModifier = new PhotoCropModifier(mPhoto, frameSize, aRect, DEFAULT_JPEG_COMPRESSION_QUALITY);
```
a-szotyori, line 92: [Comment] Awesome! 👏

a-szotyori (approved):

PR #76, opened 2022-07-25
Pia 1298 remove bolts framework

llevente:

PR #104, opened 2022-10-23
Pia 3154 update default network service

a-szotyori:
Please review after #76


a-szotyori:
Changes are required for checks in PR #76 to succeed. Will use that PR's branch for ticket PIA-3154.

PR #105, opened 2022-10-28
fix(capture-sdk): Reduce minimum aspect ratio back to 4:3

a-szotyori:
With the minimum set to 16:9 we excluded devices that we supported before. These devices are still popular (like Galaxy Tab S2) so we need to reduce the minimum back to 4:3.

PIA-3248


AdnanYupi (approved):
Looks good to me.

PR #107, opened 2022-11-03
Pia 2980 only qr scannig

AdnanYupi:


capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragmentImpl.java
```diff
@@ -657,6 +659,19 @@ private void startHelpActivity() {
...
+            mPaneWrapper.setVisibility(View.GONE);
+        }
+    }
+
+    private boolean isOnlyQRCodeScanning() {
+        if (!GiniCapture.hasInstance()) {
+            return false;
+        }
+
+        return GiniCapture.getInstance().isQRCodeScanningEnabled();
```
a-szotyori, line 673: [Request] You need to add a new method to the `GiniCapture.Builder` to enable only QR code scanning mode. 

Then `isOnlyQRCodeScanning()` should return true if QR code scanning AND QR code only mode is enabled.

capture-sdk/sdk/src/main/res/drawable/gc_only_qr_code_scanning_frame.xml
```diff
@@ -0,0 +1,26 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
a-szotyori, line 1: [Clarification] Where is this used? When I run the example app, then it shows the normal frame.

a-szotyori (changes requested):
Some more work is needed 😉.

capture-sdk/sdk/src/main/res/drawable/gc_only_qr_code_scanning_frame.xml
AdnanYupi: Oh yeah, that can be deleted it's not needed anymore. Will delete that.

a-szotyori: But the design shows a square for QR code only mode. So we do need it. Let's talk about it tomorrow.

a-szotyori: [Scan](https://www.figma.com/file/LJT8L5kpqe00Zbb2NDK55B/Gini-Photo-Payment-%2F-Android?node-id=2458%3A11788)

capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragmentImpl.java
```diff
@@ -660,17 +661,21 @@ private void startHelpActivity() {
     }
 
     private void initOnlyQRScanning() {
-        if (isOnlyQRCodeScanning()) {
+        if (isOnlyQRCodeScanningEnabled()) {
```
a-szotyori, line 664: [Request] If qr code scanning is not enabled, then this should be skipped as well. Otherwise if I set only `setOnlyQRCodeScanning(true)` then it will change the UI, but QR code scanning won't work.

a-szotyori (changes requested):

capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragmentImpl.java
AdnanYupi: oh okay

a-szotyori: If users do this configuration, then the qr code only mode should be ignored.
Screenshot 2022-11-03 at 18 49 31

AdnanYupi: Done

a-szotyori (approved):

PR #108, opened 2022-11-07
Pia 2964 review screen new UI

AdnanYupi:


capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/cache/PhotoMemoryCache.java
```diff
@@ -28,6 +28,7 @@ public class PhotoMemoryCache extends MemoryCache<ImageDocument, Photo> {
 
     private static final int RUNNING_WORKERS_LIMIT = 3;
     private final DocumentDataMemoryCache mDocumentDataMemoryCache;
+    private boolean mRotate;
```
a-szotyori, line 31: [Suggestion] Can it be deleted?

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/camera/photo/PhotoCropModifier.java
```diff
@@ -52,16 +52,17 @@ public void modify() {
                 int width1 = rotatedBitmap.getWidth() * aRect.width() / mScreenSize[0];
                 int height1 = rotatedBitmap.getHeight() * aRect.height() / mScreenSize[1];
 
-                matrix.invert(matrix);
+                //matrix.invert(matrix);
```
a-szotyori, line 55: [Request] Can be deleted.

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/camera/photo/PhotoFactoryDocumentAsyncTask.java
```diff
@@ -1,12 +1,16 @@
 package net.gini.android.capture.internal.camera.photo;
 
+import android.graphics.Bitmap;
```
a-szotyori, line 3: [Suggestion] Are these imports still used? If not they can be deleted.

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/util/ActivityHelper.java
```diff
@@ -77,6 +80,18 @@ public void handleOnBackPressed() {
         });
     }
 
+    public static Point getRealScreenSize(Activity activity) {
```
a-szotyori, line 83: [Suggestion] Can't find where this is used. Can it be deleted?

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/util/AndroidHelper.java
```diff
@@ -12,6 +12,13 @@
  */
 public final class AndroidHelper {
 
+    //We need this only while app is running
+    //Shared preferences would be maybe overthinking
+    //Let me know if we don't want to use this
+
+    public static int STORE_SCROLL_STATE = -1;
+    public static int STORE_DOCUMENTS_SIZE = -1;
```
a-szotyori, line 20: [Suggestion] I would propose to not store any global state. It would make it tough to rediscover later how state is managed and it also couples state management of the review screen to other classes (for example the CameraActivity). Also debugging can be more difficult with global states.

Instead I would propose to do the following:
* Camera screen knows when an image was added and tells the review screen whether to scroll to the last page or not: the `MultiPageReviewActivity.createIntent()` receives a boolean parameter called `shouldScrollToLastPage` and forwards this to `MultiPageReviewFragment.createInstance()` where it can be passed to the fragment via an attribute.
* Remember which page was shown when returning to the review screen: using `STORE_SCROLL_STATE` is ok for now. We will have to [change navigation so that the review screens becomes root after taking a picture](https://ginis.atlassian.net/browse/PIA-3146) so I am fairly sure we will be able to drop `STORE_SCROLL_STATE` then.

a-szotyori (changes requested):
Your ViewPager "kung-fu" is astonishing! There are some other things I found though 👀

a-szotyori (approved):
Looks good! Thank you!

PR #109, opened 2022-11-08
Pia 2990 no results UI

llevente:


capture-sdk/sdk/src/main/res/drawable/ic_alert_triangle.xml
```diff
@@ -0,0 +1,27 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
a-szotyori, line 1: [Request] We need to use the `gc_` prefix for all resource names.

capture-sdk/sdk/src/main/res/values/dimens.xml
```diff
@@ -100,25 +100,27 @@
...
-    <dimen name="gc_noresults_tip_text_bottom_margin">16dp</dimen>
+    <dimen name="gc_noresults_tip_text_bottom_margin">12dp</dimen>
     <dimen name="gc_noresults_tip_text_right_margin">16dp</dimen>
     <dimen name="gc_noresults_tip_text_left_margin">8dp</dimen>
     <dimen name="gc_noresults_tip_text_max_width">250dp</dimen>
     <dimen name="gc_noresults_space_height_for_bottom_button">64dp</dimen>
-    <dimen name="gc_noresults_bottom_button_height">48dp</dimen>
-    <dimen name="gc_noresults_bottom_button_margin">16dp</dimen>
-    <dimen name="gc_noresults_bottom_button_max_width">300dp</dimen>
+    <dimen name="gc_noresults_bottom_button_margin">21dp</dimen>
```
a-szotyori, line 114: [Request] We should try to use "size classes" we have like `small` or `large` for margins. Instead of this custom dimen we should use `large_24`. If somebody notices the 3dp difference then send them to me 😄

capture-sdk/sdk/src/main/res/values/dimens.xml
```diff
@@ -100,25 +100,27 @@
...
+    <dimen name="gc_noresults_bottom_button_max_width">312dp</dimen>
+    <dimen name="gc_noresults_bottom_button_top_margin">12dp</dimen>
```
a-szotyori, line 116: [Request] Similar to above: please use the `medium` dimen instead.

a-szotyori (changes requested):
Nice work! 👍 I noticed only three small things.

PR #114, opened 2022-11-10
Pia 2311 no result bottom navbar

llevente:


capture-sdk/sdk/src/main/java/net/gini/android/capture/GiniCapture.java
```diff
@@ -1092,6 +1102,9 @@ public CameraNavigationBarBottomAdapter getCameraNavigationBarBottomAdapter() {
             return cameraNavigationBarBottomAdapter;
         }
 
+        public NoResultsNavigationBarBottomAdapter getNoResultsNavigationBarBottomAdapter() {
```
a-szotyori, line 1105: [Request] We also need a setter for this like we do for the other bottom navigation bars (e.g., `setOnboardingNavigationBarBottomAdapter()`).

a-szotyori (changes requested):
Nice! We just need a setter for the new bottom navbar adapter.

a-szotyori (approved):

PR #116, opened 2022-11-13
fix(capture-sdk): Unknown qr code updated

AdnanYupi:
PIA-3279

Unknown QR code UI updated and logic of showing. 
To remove the flashing, between each show there is a one-second delay.

### DIFF
--- a/capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragmentImpl.java
+++ b/capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragmentImpl.java
@@ -11,6 +11,8 @@ import android.graphics.Rect;
 import android.media.Image;
 import android.net.Uri;
 import android.os.Bundle;
+import android.os.Handler;
+import android.os.Looper;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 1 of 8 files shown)


capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragmentImpl.java
```diff
@@ -222,11 +224,21 @@ private void handleQRCodeDetected(@Nullable final PaymentQRCodeData paymentQRCod
...
-            mPaymentQRCodePopup.show(paymentQRCodeData);
-        }
+        if (mUnsupportedQRCodePopup.isShown() || mPaymentQRCodePopup.isShown())
+            return;
+
+        showQRCodeView(paymentQRCodeData);
+    }
+
+    private void showQRCodeView(PaymentQRCodeData data) {
+        new Handler(Looper.getMainLooper())
```
a-szotyori, line 234: [Request] Two things I'm missing here to fulfill the requirements in the ticket ([see the first bullet point](https://ginis.atlassian.net/browse/PIA-3279)):
* The delay should only apply to the same QR code. If the QR code changes, then the delay should be ignored.
  When I tested there was a delay before showing the QR code UI for the new QR code.
* When detecting a QR code for the first time, then the delay should be ignored.
  When I tested there was a delay before showing the QR code UI for the first QR code.

a-szotyori (changes requested):
Great progress! Some logic refinement and tablet layouts remaining.

capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragmentImpl.java
AdnanYupi: Ah shit I totally forgot tablets

a-szotyori (approved):
Thank you for the changes!

PR #120, opened 2022-11-15
Pia 3234 camera injected loading indicator

AdnanYupi:
Pia 3234

Custom loading indicator injected in the camera. There was a possible issue with an injected view not working properly, with the progress bar showing only on the first shot, workaround is to inflate the view again in onResume


a-szotyori (approved):
Nice! Shoutout to @llevente for his help!

PR #121, opened 2022-11-15
Pia 3151 permission denied UI

AdnanYupi:
PIA 3151 

No camera permission UI updated.
Added an option on Home Screen to overpass camera permission.


a-szotyori (changes requested):
Nice! One little change is needed so that the top and bottom navigation bars remain visible.

a-szotyori (approved):
Thank you for the change!

PR #128, opened 2022-11-21
fix(capture-sdk): changed style name

AdnanYupi:
- margin fixed
- changed style name and TextView style added

PIA-2967


a-szotyori (approved):
Nice! PR #129 will have to be updated to use `Caption1` as well, right?

PR #136, opened 2022-11-29
fix(capture-sdk): PIA-3362

AdnanYupi:
- Fixed German and English localization
- Changed orders of help items to match design and iOS


a-szotyori (approved):
Thank you!

PR #138, opened 2022-12-01
feat(capture-sdk): 

AdnanYupi:
- the bottom bar was removed in only QR code mode
- added missing string scan

PIA-3373


a-szotyori (approved):
Nice!

PR #144, opened 2022-12-02
fix: Add proguard rule to libraries to preserve all retrofit service interface methods

a-szotyori:
Affects: core-api-library, bank-api-library, health-api-library

PIA-3387


AdnanYupi (approved):
(Y)

PR #146, opened 2022-12-06
Pia 2500 refactor feedback sending

AdnanYupi:
- feedback sending refactored 

PIA-2500


a-szotyori (approved):
Awesome!

capture-sdk/sdk/src/main/java/net/gini/android/capture/Amount.java
```diff
@@ -0,0 +1,35 @@
+package net.gini.android.capture;
+
+import androidx.annotation.NonNull;
+
+import java.math.BigDecimal;
+
+public class Amount {
+
+    private static Amount mAmount;
```
a-szotyori, line 9: [Suggestion] What about simply creating a public static final property like this (or a companion object property with `@JvmStatic` annotation in Kotlin):
```
public static final Amount EMPTY = new Amount(BigDecimal.valueOf(0), AmountCurrency.EUR);
```

This way it's less code and also [thread safe](https://stackoverflow.com/a/25868299).

capture-sdk/sdk/src/main/java/net/gini/android/capture/Amount.java
```diff
@@ -0,0 +1,35 @@
...
+
+
+
+    public Amount(@NonNull final BigDecimal decimal, @NonNull final AmountCurrency currency) {
+        mBigDecimal = decimal;
+        mCurrency = currency;
+    }
+
+
+    public String amountToPay() {
```
a-szotyori, line 25: [Request] I would prefer to avoid using double and instead use a DecimalFormat to format the number. Something like this:
```
final DecimalFormat decimalFormat = new DecimalFormat("0.00", DecimalFormatSymbols.getInstance(Locale.ENGLISH));
decimalFormat.setParseBigDecimal(true);
return decimalFormat.format(mBigDecimal) + ":" + mCurrency.name();
```

capture-sdk/sdk/src/main/java/net/gini/android/capture/Amount.java
```diff
@@ -0,0 +1,35 @@
+package net.gini.android.capture;
+
+import androidx.annotation.NonNull;
+
+import java.math.BigDecimal;
+
+public class Amount {
```
a-szotyori, line 7: [Request] Can you please rewrite this into Kotlin?

a-szotyori (changes requested):
Almost there 😁

capture-sdk/sdk/src/main/java/net/gini/android/capture/Amount.java
AdnanYupi: Okay totally makes sense. Thanks

a-szotyori (approved):
Thank you!

capture-sdk/sdk/src/main/java/net/gini/android/capture/Amount.kt
```diff
@@ -5,17 +5,36 @@ import java.text.DecimalFormat
...
+     * @suppress
+     */
     fun amountToPay(): String {
         val decimalFormat = DecimalFormat("0.00", DecimalFormatSymbols.getInstance(Locale.ENGLISH))
         decimalFormat.isParseBigDecimal = true
         return decimalFormat.format(mBigDecimal) + ":" + mCurrency.name
     }
 
+    /**
+     * Creates thread safe static instance of Amount class
```
a-szotyori, line 36: [Request] This comment should be moved to be above `@JvmStatic EMPTY`, otherwise the generated dokka will show the comment for the companion and not for `EMPTY`.

capture-sdk/sdk/src/main/java/net/gini/android/capture/AmountCurrency.kt
```diff
@@ -1,5 +1,18 @@
...
+ *
+ * <p>
+ * Currency codes expressed in ISO 4217 standard </p>
+ *
+ * <ul>
+ *  <li> EUR - stands for EURO (official currency of the euro-zone)
+ *  <li> GBP - stands for the British pound sterling
+ *  <li> USD - stands for United States Dollar
+ *  <li> CHF - stands for Confoederatio Helvetica franc
+ * </ul>
```
a-szotyori, line 14: [Request] Please add dots to the end of the sentences.

Also you can use markdown with kdoc so we can get rid of html tags.

a-szotyori (changes requested):
Some changes needed. 

You can check how the generated dokka reference documentation looks by running `./gradlew capture-sdk:sdk:dokkaHtmlSiblingCollector` and then open `capture-sdk/sdk/build/docs/dokka/index.html` in the browser.

capture-sdk/sdk/src/main/java/net/gini/android/capture/GiniCapture.java
```diff
@@ -171,20 +179,79 @@ public static synchronized Builder newInstance() {
...
-            if (sInstance.mNetworkRequestsManager != null) {
-                sInstance.mNetworkRequestsManager.cleanup();
-            }
-            if (sInstance.mGiniCaptureNetworkApi != null) {
-                sInstance.mGiniCaptureNetworkApi.setUpdatedCompoundExtractions(Collections.<String, GiniCaptureCompoundExtraction>emptyMap());
-            }
-            sInstance.mImageMultiPageDocumentMemoryStore.clear();
-            sInstance.internal().setReviewScreenAnalysisError(null);
-            sInstance = null; // NOPMD
+    public static synchronized void cleanup(@NonNull final Context context,
```
a-szotyori, line 182: [Request] We also updated this method and we should document what's the purpose for the new params and also list them with `@param`.

a-szotyori (changes requested):

PR #154, opened 2022-12-12
feat(capture-sdk):

AdnanYupi:
- Created color in attr
- Color override in light and night styles to match design

PIA-3319


a-szotyori (approved):
Nice!

PR #156, opened 2022-12-15
feat(capture-sdk):

AdnanYupi:
- reference documentation for only QR code scanning

PIA-2978


a-szotyori (changes requested):
Sorry, Adnan, I overlooked something.

Can you please also add this configuration option to `CaptureConfiguration` in Bank SDK? Thank you!

PR #157, opened 2022-12-15
docs(capture-sdk): Updated integration guide for Help

llevente:


capture-sdk/sdk/src/doc/source/features.rst
```diff
@@ -381,6 +383,18 @@ You can show your own help screens in the Gini Capture SDK. You can pass the tit
 You can also disable the supported formats help screen by passing ``false`` to
 ``GiniCapture.Builder.setSupportedFormatsHelpScreenEnabled()``.
 
+Furthermore you can set back navigation action on a custom view on bottom navigation bar. You can pass your custom ``HelpNavigationBarBottomAdapter`` implementation to
```
a-szotyori, line 386: [Suggestion] When I read `set back navigation action` then I think I can set an action to be triggered on back navigation. Whereas in reality I can show a button to which I have to add the back action provided by the SDK. 

What do you think about using `show back navigation button` instead: `Furthermoe you can show a back navigation button on a custom view ...`

capture-sdk/sdk/src/main/java/net/gini/android/capture/help/view/HelpNavigationBarBottomAdapter.kt
```diff
@@ -7,11 +7,16 @@ import net.gini.android.capture.databinding.GcHelpNavigationBarBottomBinding
...
 /**
- * Implement this interface if you would like custom action triggered when tapping back navigation
+ * Implement this interface to add back navigation to a custom view on bottom navigation bar
  * and pass it to the {@link GiniCapture.Builder#setHelpNavigationBarBottomAdapter(HelpNavigationBarBottomAdapter)}
  */
 interface HelpNavigationBarBottomAdapter : InjectedViewAdapter {
 
+    /**
+     * <p>
+     *  Set a listener for back navigation on a view on bottom navigation bar layout.
```
a-szotyori, line 17: [Suggestion] I would suggest to only say here `Set a listener for the back navigation button.` 

I don't think we have to repeat `bottom navigation bar layout`. I find it's implied by the context. But I have no strong oppionin. Feel free to keep it. It's better to have more information than less.

a-szotyori (commented):
Just some suggestions to discuss.

PR #158, opened 2022-12-15
Pia 2913 help migration guide

llevente:


a-szotyori (approved):
I see it builds upon PR #157 so please merge it after that one.

PR #169, opened 2022-12-28
feat(capture-sdk):

AdnanYupi:
- Camera reference documentation

PIA-2878

[PIA-2878]: https://ginis.atlassian.net/browse/PIA-2878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragmentImpl.java
```diff
@@ -464,11 +471,6 @@ public void onInterrupted() {
         });
     }
 
-    @VisibleForTesting
```
zladzeyka, line 467: [Clarification] why did we remove it? Don't we use it for tests?

capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragmentListener.java
```diff
@@ -31,6 +31,16 @@ interface CameraFragmentListener {
...
     void onDocumentAvailable(@NonNull Document document);
 
+
+    /**
+     * <p>
+     * Called when starting MultiPageReview Screen
+     * </p>
+     *
+     * @param multiPageDocument the image taken by the camera or the validated imported document
+     * @param shouldScrollToLastPage if pages should scroll to the last one
```
zladzeyka, line 41: [Suggestion] Maybe it is worth mentioning that `shouldScrollToLastPage` is a flag that defines if pages should scroll to the last one?

zladzeyka (changes requested):
Looks good.
I'm a bit surprised to see a small refactoring in the reference documentation update 🙈.
I have only a small comment.

capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragmentImpl.java
AdnanYupi: @zladzeyka I don't think we used this line of code anymore even in test cases:
` @VisibleForTesting
    PaymentQRCodeReader getPaymentQRCodeReader() {
        return mPaymentQRCodeReader;
    }`

a-szotyori (approved):
Thanks for also switching to lambdas! 👍

PR #172, opened 2023-01-09
Pia 3496 fix fastlane in GitHub actions

a-szotyori:
* Update fastlane action and setup ruby in a dedicated action.
* Update all GitHub Actions versions.
* Trigger workflows on pull request only if the event type is opened, edited or reopened.


AdnanYupi (approved):
Thanks for fixing this!

PR #186, opened 2023-01-18
docs(capture-sdk): Added Error screen integration doc

llevente:


capture-sdk/sdk/src/doc/source/features.rst
```diff
@@ -421,6 +421,20 @@ You can show customized activity indicator on this screen. You can pass your cus
...
+``GiniCapture.Builder``:
+
+.. code-block:: java
+
+    CustomErrorNavigationBarBottomAdapter customErrorNavigationBarBottomAdapter = new CustomErrorNavigationBarBottomAdapter();
+
+    GiniCapture.newInstance()
+            .setErrorNavigationBarBottomAdapter(customErrorNavigationBarBottomAdapter)
+            .build();
+
```
a-szotyori, line 437: [Request] We should also mention that they should handle the `CameraActivity.RESULT_ENTER_MANUALLY` result code which is returned by the SDK when the user clicks the "Enter manually" button on the error screen.

a-szotyori (changes requested):
We should inform clients about the new `CameraActivity.RESULT_ENTER_MANUALLY` result code.

a-szotyori (approved):
Thank you!

PR #192, opened 2023-01-24
docs(capture-sdk): Updated reference doc for No results

llevente:


capture-sdk/sdk/src/main/java/net/gini/android/capture/noresults/NoResultsActivity.java
```diff
@@ -43,96 +43,40 @@
  *     <ul>
  *         <li>
  *             <b>Header icon:</b> via images for mdpi, hdpi, xhdpi, xxhdpi, xxxhdpi named
- *             {@code gc_alert_icon.png}
```
a-szotyori, line 46: [Request] Can you please add `Internal use only.` at the top of the javadoc for `NoResultsActivity`? In the new version we don't want clients to use the activity directly anymore.

a-szotyori (changes requested):
Just a small request. Looks good otherwise!

capture-sdk/sdk/src/main/java/net/gini/android/capture/noresults/NoResultsActivity.java
```diff
@@ -25,6 +25,10 @@
 import static net.gini.android.capture.camera.CameraActivity.RESULT_ENTER_MANUALLY;
 import static net.gini.android.capture.internal.util.ActivityHelper.interceptOnBackPressed;
 
+/**
+ * Internal use only
+ */
+
 /**
```
a-szotyori, line 32: [Request] Sorry, Lenci, for annoying you. Please merge these two javadoc comment blocks so that `Internal use only.` is shown in the generated reference documentation. Thank you!

P.S.: Don't forget the dot from the end of the sentence. 😉

a-szotyori (changes requested):
One small change for you, but a giant leap for our documentation. 👨‍🚀

a-szotyori (approved):
Thank you!

PR #201, opened 2023-01-27
feat(bank-sdk):

AdnanYupi:
- Injecting image container

PIA-3580

### DIFF
--- a/bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/GiniBank.kt
+++ b/bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/GiniBank.kt
@@ -18,7 +18,9 @@ import net.gini.android.bank.sdk.capture.CaptureConfiguration
 import net.gini.android.bank.sdk.capture.CaptureImportInput
 import net.gini.android.bank.sdk.capture.applyConfiguration
 import net.gini.android.bank.sdk.capture.digitalinvoice.view.DefaultDigitalInvoiceOnboardingNavigationBarBottomAdapter
+import net.gini.android.bank.sdk.capture.digitalinvoice.view.DigitalInvoiceOnboardingIllustrationAdapter
 import net.gini.android.bank.sdk.capture.digitalinvoice.view.DigitalInvoiceOnboardingNavigationBarBottomAdapter
+import net.gini.android.bank.sdk.capture.digitalinvoice.view.ImageDigitalInvoiceOnboardingIllustrationAdapter
 import net.gini.android.bank.sdk.capture.util.getImportFileCallback
 import net.gini.android.bank.sdk.error.AmountParsingException
 import net.gini.android.bank.sdk.pay.getBusinessIntent
@@ -54,7 +56,7 @@ object GiniBank {
     private var giniApi: GiniBankAPI? = null
 
     var digitalInvoiceOnboardingNavigationBarBottomAdapter: DigitalInvoiceOnboardingNavigationBarBottomAdapter = DefaultDigitalInvoiceOnboardingNavigationBarBottomAdapter()
-
+    var digitalInvoiceOnboardingIllustrationAdapter: DigitalInvoiceOnboardingIllustrationAdapter = ImageDigitalInvoiceOnboardingIllustrationAdapter()
     internal fun getCaptureConfiguration() = captureConfiguration
 
     /**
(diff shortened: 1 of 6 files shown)


bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/capture/digitalinvoice/view/DigitalInvoiceOnboardingIllustrationAdapter.kt
```diff
@@ -0,0 +1,43 @@
...
+
+import android.view.View
+import android.view.ViewGroup
+import android.widget.ImageView
+import androidx.annotation.DrawableRes
+import androidx.annotation.StringRes
+import androidx.core.content.ContextCompat
+import net.gini.android.capture.view.InjectedViewAdapter
+
+interface DigitalInvoiceOnboardingIllustrationAdapter: InjectedViewAdapter {
```
a-szotyori, line 11: [Suggestion] I see our `OnboardingIllustrationAdapter` in Capture SDK is generic enough to be used in the Bank SDK as well. 

What about making it public and reusing it? Then we don't need a dedicated one for digital invoice.

bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/capture/digitalinvoice/view/DigitalInvoiceOnboardingIllustrationAdapter.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    fun onVisible()
+
+    fun onHidden()
+
+    fun onIllustrationSet(@DrawableRes drawableRes: Int)
+}
+
+class ImageDigitalInvoiceOnboardingIllustrationAdapter(): DigitalInvoiceOnboardingIllustrationAdapter {
```
a-szotyori, line 20: [Suggestion] Similarly to the above comment, the `ImageOnboardingIllustrationAdapter` is generic enough to be used in the Bank SDK as well.

If we make it public and reuse it, then we can reduce the amount of code.

bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/GiniBank.kt
```diff
@@ -54,7 +56,7 @@ object GiniBank {
     private var giniApi: GiniBankAPI? = null
 
     var digitalInvoiceOnboardingNavigationBarBottomAdapter: DigitalInvoiceOnboardingNavigationBarBottomAdapter = DefaultDigitalInvoiceOnboardingNavigationBarBottomAdapter()
-
+    var digitalInvoiceOnboardingIllustrationAdapter: DigitalInvoiceOnboardingIllustrationAdapter = ImageDigitalInvoiceOnboardingIllustrationAdapter()
```
a-szotyori, line 59: [Suggestion] If we make `OnboardingIllustrationAdapter` and `ImageOnboardingIllustrationAdapter` public, then we could reuse them in Bank SDK and can avoid some duplication. What do you think?
```suggestion
    var digitalInvoiceOnboardingIllustrationAdapter: OnboardingIllustrationAdapter = ImageOnboardingIllustrationAdapter(R.drawable.gbs_digital_invoice_list_image, R.string.gbs_digital_invoice_list_image_content_description)
```

Oh and we also found the content description for the image! 🙈 Please create the string resources for it and use `Digital invoice illustration` as the value for both DE and EN for now. Thank you!

a-szotyori (changes requested):
What do you think about reusing stuff from Capture SDK?

bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/capture/digitalinvoice/onboarding/DigitalInvoiceOnboardingFragment.kt
```diff
@@ -130,17 +131,26 @@ class DigitalInvoiceOnboardingFragment : Fragment(), DigitalOnboardingScreenCont
...
         setInputHandlers()
+        setupImageIllustrationAdapter()
         setupOnboardingBottomNavigationBar()
     }
 
 
+    private fun setupImageIllustrationAdapter() {
+        if (GiniCapture.hasInstance()) {
+            binding.digitalInvoiceImageContainer.injectedViewAdapter =
+                GiniBank.digitalInvoiceOnboardingIllustrationAdapter
```
a-szotyori, line 142: [Request] We also need to call `onVisible` when the fragment starts and `onHidden` when the fragment stops.

a-szotyori (changes requested):
Can you please also add a custom illustration example using Lotti like we have in the Capture SDK example app? Thank you!

a-szotyori (approved):
Nice! Love the lottie animation you picked!

PR #202, opened 2023-01-27
fix(capture-sdk): Use string resources for button titles

a-szotyori:
PIA-3661


llevente (approved):

a-szotyori:
Merged to include it into the release candidate we create today.

PR #206, opened 2023-01-30
fix(capture-sdk): Fix analysis cancellation handling

a-szotyori:
Don't show the error screen when analysis was cancelled.

PIA-3666


llevente (approved):
It is working and looks good

PR #215, opened 2023-02-03
fix(capture-sdk):

AdnanYupi:
- Crash on fast double click on delete page

PIA-3695


capture-sdk/sdk/src/main/java/net/gini/android/capture/ClickListenerExt.kt
```diff
@@ -0,0 +1,7 @@
+package net.gini.android.capture
+
+import android.view.View
+
+fun View.onClick(click: (View) -> Unit) {
```
a-szotyori, line 5: [Request] Let's make it clear what this method does by renaming it to `View.setIntervalClickListener`.

a-szotyori (changes requested):
Just some small things.

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/ui/IntervalClickListener.kt
```diff
@@ -0,0 +1,23 @@
...
+
+    override fun onClick(view: View) {
+        if (isEnabled) {
+            isEnabled = false
+            view.postDelayed(enable, interval)
+            click.onClick(view)
+        }
+    }
+
+    var isEnabled = true
```
a-szotyori, line 20: [Request] This should be private as well.

a-szotyori (changes requested):
One small thing and then it's good to go.

a-szotyori (approved):
Thank you!

PR #218, opened 2023-02-07
Bottom nav bar

AdnanYupi:
- Navigation bottom bar added for Digital Invoice

PIA 3570


bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/capture/digitalinvoice/view/DigitalInvoiceNavigationBarBottomAdapter.kt
```diff
@@ -0,0 +1,65 @@
...
+
+interface DigitalInvoiceNavigationBarBottomAdapter : InjectedViewAdapter {
+
+    fun setOnBackClickListener(listener: OnClickListener)
+
+    fun setOnHelpClickListener(listener: OnClickListener)
+
+    fun setOnPayClickListener(listener: OnClickListener)
+
+    fun setHelpButtonResource(@DrawableRes drawable: Int)
```
a-szotyori, line 19: [Request] We don't need this method. It's up to the implementation what drawable (if any) will be shown. In our default implementation we'll use our drawable, but clients could for example show a button with only text.

a-szotyori (changes requested):

a-szotyori (commented):
Thank you, the code looks good now! I found some UI issues that need to be fixed before I can approve the PR.

a-szotyori (changes requested):
Besides the comment I also saw that on phones the "Total" and the amount are still aligned to with the button instead of the screen edge:
Screenshot_1676024997

bank-sdk/sdk/src/main/res/layout/gbs_digital_invoice_navigation_bar_bottom.xml
```diff
@@ -85,14 +87,14 @@
         android:id="@+id/gbs_back_btn"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
-        android:layout_marginStart="@dimen/medium"
+        android:layout_marginStart="8dp"
```
a-szotyori, line 90: [Request] `@dimen/medium`

a-szotyori (changes requested):
One small fix left.

a-szotyori (approved):
Yeeah! 

goal

PR #221, opened 2023-02-10
Pia 3694 fast clicks duplicate screens

AdnanYupi:
- refactored current code

PIA-3694


bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/capture/digitalinvoice/help/view/DigitalInvoiceHelpNavigationBarBottomAdapter.kt
```diff
@@ -31,7 +32,7 @@ class DefaultDigitalInvoiceHelpNavigationBarBottomAdapter: DigitalInvoiceHelpNav
     var viewBinding: GbsHelpNavigationBarBottomBinding? = null
 
     override fun setOnBackButtonClickListener(listener: View.OnClickListener) {
-        viewBinding?.gbsGoBack?.setOnClickListener(listener)
+        viewBinding?.gbsGoBack?.setIntervalClickListener(listener)
```
a-szotyori, line 35: [Suggestion] Can't we pass in an instance of `IntervalClickListener`? It is an implementation of `View.OnClickListener` and we could prevent quick double tap issues also for custom adapters.

Instead of this:
`adapter.setOnBackButtonClickListener() { v: View? -> onBackPressed() }`

We could do this:
`adapter.setOnBackButtonClickListener( IntervalClickListener { v: View? -> onBackPressed() })`

Then we don't have to modify the default adapter implementations. If you think this could work, then it would be great if you could apply this change for all our view adapters.

capture-sdk/sdk/src/main/java/net/gini/android/capture/view/NavigationBarTopAdapter.kt
```diff
@@ -90,7 +91,7 @@ class DefaultNavigationBarTopAdapter : NavigationBarTopAdapter {
         if (GiniCapture.hasInstance()
             && GiniCapture.getInstance().isBottomNavigationBarEnabled
         ) {
-            viewBinding?.gcNavigationBar?.setOnMenuItemClickListener {
+            viewBinding?.gcNavigationBar?.setOnMenuItemIntervalClickListener {
```
a-szotyori, line 94: [Suggestion] Same as in my previous comment. If the listener we pass in is `IntervalClickListener` then we don't have to use `setOnMenuItemIntervalClickListener`.

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/ui/IntervalToolbarMenuItemIntervalClickListener.kt
```diff
@@ -0,0 +1,24 @@
+package net.gini.android.capture.internal.ui
+
+import android.os.Handler
+import android.os.Looper
+import android.view.MenuItem
+import androidx.appcompat.widget.Toolbar
+
+class IntervalToolbarMenuItemIntervalClickListener(private val itemListener: Toolbar.OnMenuItemClickListener?): Toolbar.OnMenuItemClickListener,
```
a-szotyori, line 8: [Suggestion] Similarly to my other comments we should pass an instance of this to the `NavigationBarTopAdapter.setOnMenuItemClickListener()`.

a-szotyori (commented):
I left some suggestions. I'm curious what you think about them. 🤓

a-szotyori (approved):
I pushed a commit to revert using the default click listener interfaces in the adapter interfaces. Clients don't need to know which exact implementation of the click listener is passed to the adapter. 

This also prevents our public API to be bound to the `IntervalClickListener` so that we are free to exchange it with any other `View.OnClickListener` implementation later.

PR #226, opened 2023-02-14
fix(capture-sdk):

AdnanYupi:
- fixed bottom click not working if snackbar is presented

PIA-3772


a-szotyori (approved):
Thanks! This one should go into QA to be tested on different devices and Android versions.

PR #227, opened 2023-02-14
fix(capture-sdk):

AdnanYupi:
- strings changed

PIA-3777


a-szotyori (approved):
Thanks, you can also merge it. No need to QA this simple change.

PR #235, opened 2023-02-16
fix(bank-sdk):

AdnanYupi:
- onboarding scrollable 

Pia 3576


a-szotyori (approved):
Thank you, looks good now!

PR #236, opened 2023-02-16
feat(capture-sdk):

AdnanYupi:
- useful tips

PIA-3773


a-szotyori (changes requested):
Looks good but the "Supported formats" header is shown twice now.

a-szotyori (approved):
Looks good now, thank you!

PR #239, opened 2023-02-17
Pia 3593 ra detail reference doc

llevente:


bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/capture/digitalinvoice/details/LineItemDetailsInterface.kt
```diff
@@ -7,14 +7,14 @@ package net.gini.android.bank.sdk.capture.digitalinvoice.details
  */
 
 /**
- * Public API of the [LineItemDetailsFragment].
+ * Public API of the [DigitalInvoiceBottomSheet].
  *
  */
-interface LineItemDetailsFragmentInterface {
+interface LineItemDetailsInterface {
```
a-szotyori, line 13: [Comment] Thank you for renaming the interfaces!

a-szotyori, line 13: [Request] We can remove these interfaces from the SDK's public API. We don't want clients to be able to implement these anymore. You can make them `internal`. If that's not possible, then just add `Internal use only.`. Thank you!

bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/capture/digitalinvoice/details/LineItemDetailsListener.kt
```diff
@@ -9,9 +9,9 @@ import net.gini.android.bank.sdk.capture.digitalinvoice.SelectableLineItem
  */
 
 /**
- * Interface used by the [LineItemDetailsFragment] to dispatch events to the hosting Activity.
+ * Interface used by the [DigitalInvoiceActivity] to dispatch events to it's fragment.
  */
-interface LineItemDetailsFragmentListener {
+interface LineItemDetailsListener {
```
a-szotyori, line 14: [Request] Can be `internal` as well or add `Internal use only.`.

a-szotyori (changes requested):
Looks good, just the interfaces need to be made internal now.

bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/capture/digitalinvoice/details/LineItemDetailsListener.kt
llevente: [Note]: Made interfaces internal but then had to make the "internal use only" DigitalInvoiceBottomSheet class too, because it exposed the variable from the interface.  Please check if that is good or the listener interface should remain public with the Internal comment...

a-szotyori: That is great, thanks!

a-szotyori (approved):

PR #241, opened 2023-02-20
feat(capture-sdk):

AdnanYupi:
- add haptic feedback on QR code scanned

PIA-3500

### DIFF
--- a/capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/camera/view/QRCodePopup.kt
+++ b/capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/camera/view/QRCodePopup.kt
@@ -3,6 +3,7 @@ package net.gini.android.capture.internal.camera.view
 import android.content.res.ColorStateList
 import android.os.Handler
 import android.os.Looper
+import android.view.HapticFeedbackConstants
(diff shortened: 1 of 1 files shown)


capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/camera/view/QRCodePopup.kt
```diff
@@ -113,7 +115,9 @@ internal class QRCodePopup<T> @JvmOverloads constructor(
                     R.color.Success_05
                 )
             )
+            performHapticFeedback(HapticFeedbackConstants.CONFIRM)
```
a-szotyori, line 118: [Clarification] The Android documentation says the `CONFIRM` and `REJECT` constants are available from [api level 30 (Android 11)](https://developer.android.com/reference/kotlin/android/view/HapticFeedbackConstants?hl=en#CONFIRM:kotlin.Int). Do you have a device running Android 10 or older to test it? I will test it as well, but my device is still charging.

a-szotyori (commented):
We need to check what happens when running on Android 10 or older.

a-szotyori (approved):

PR #248, opened 2023-02-23
feat(capture-sdk):

AdnanYupi:
- return reasons dialog UI update

PIA-3795


a-szotyori (changes requested):
The only issue is that the color of the title should be different in light and dark modes from the other items:
Screenshot_1677146751

Screenshot_1677146756

a-szotyori (approved):
Thank you! Please merge it and we'll QA it in the release candidate.

PR #249, opened 2023-02-24
feat(capture:sdk):

AdnanYupi:
- navbutton type none


a-szotyori (approved):
Looks good! Can go into QA (this will be part of the RC in two weeks).

PR #250, opened 2023-02-24
Pia 3765 release candidate crash

llevente:


a-szotyori (commented):
Looks good! After you add the GiniCapture instance check to the no results screen I'll approve it.

a-szotyori (approved):
Thank you! Can go to QA.

PR #253, opened 2023-02-28
feat(bank-sdk):

AdnanYupi:
- Close button added
- Back button from bottom bar hidden

PIA-3797


a-szotyori (commented):
I made a small commit so that the top navbar alsways shows CLOSE as the navbutton and that we set the menu for the help button only if bottom navbar is disabled.

The same crash happens as on the camera screen when bottom navbar is enabled and the close button is clicked. So this PR is blocked until that is fixed.

PR #256, opened 2023-03-02
Pia 3799 fix crash when closing camera screen

a-szotyori:
Fix reuse of `InjectedViewAdapter`s in multiple `InjectedViewContainer`s.

Keep the association between an `InjectedViewAdapter` and an `InjectedViewContainer` in the new
`InjectedViewAdapterInstance` class. This is used to determine when an `InjectedViewContainer`
owns an adapter.

Associate the `InjectedViewAdapter` with the code it needs for configuration by using the new
`InjectedViewAdapterHolder`. This permits the adapter's view to be configured every time the view
is recreated.

The `InjectedViewContainer` that has entered the started state last will take ownership of the `InjectedViewAdapter`, recreate its view, and reconfigure it. When the `InjectedViewContainer` is destroyed it only destroys the `InjectedViewAdapter`'s view if it owns that adapter.


llevente (approved):
Awesome work. The end result is a logical, nice and clean solution

PR #262, opened 2023-03-09
fix(capture-sdk): Fixed help snackbar

llevente:


a-szotyori (approved):
Great work!

a-szotyori:
Oh and this one should go into QA as usual. We have code freeze for the relase candidate now.

PR #263, opened 2023-03-10
fix(bank-sdk): Fixed RA Edit sheet UI

llevente:


a-szotyori (approved):
Nice! Can go into QA.

PR #269, opened 2023-03-17
Pia 3869 update secondary button colors

a-szotyori:


llevente (approved):

a-szotyori:
I saw the UI/UX designer didn't keep the accent color on the "no camera permission" screen's button so I removed the commit which applied the accent color there.

PR #277, opened 2023-03-27
Pia 3892 old bank extractions edit

llevente:
Please check values sent in network inspector, at me now the data sent looks fine


a-szotyori (approved):
Works well. I couldn't reproduce the issue where the changes were not visible in the sent feedback.

I will investigate the failed CI jobs.

PR #280, opened 2023-03-29
Pia 3917 bank detail accessibility

llevente:
Also contains fix for PIA-3933


bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/capture/digitalinvoice/LineItemsAdapter.kt
```diff
@@ -117,7 +116,9 @@ internal class LineItemsAdapter(private val listener: LineItemsAdapterListener)
                     viewHolder.bind(Pair(it, enabled), null)
                 }
 
-
+                // Adding padding for the last addon item, so the item looks full height without modifying the layout file
+                val bottomPadding =  if (position == (lineItems.size + addons.size - 1)) context.resources.getDimension(R.dimen.gc_large).toInt() else 0
```
a-szotyori, line 120: [Suggestion] If we replace `lineItems.size + addons.size` with a call to `getItemCount()` then we preserve it as the single place where item count is calculated.

a-szotyori (changes requested):
I just have a small suggestion.

a-szotyori (approved):
Thank you!

PR #286, opened 2023-04-04
Pia 3741 remove obsolete stuff

a-szotyori:
Lots of files were removed and I made a small fix for an issue I discovered while removing the old single-page review screen.

It's also ok if you don’t go through all the changes and instead run the bank sdk and capture sdk example apps and test the basic scenarios (take pictures, delete pages, upload images and pdfs, and import images and pdfs from other apps).


llevente:
Both SDK's seems to be working as before. Only issue is the test class for the removed single page review fragment. I think it can be removed completely.

llevente (approved):

a-szotyori:
Will be QA'd in the release candidate.

PR #287, opened 2023-04-04
WIP: Pia 3928 accessibility talkbar

llevente:
[NOTE]: Contains fix for PIA-3927 also... These were not buttons, but different elements put together in a wrapper layout so the accessibility couldn't work good, because it could select the textview and the image separately and it read the label's text but did not know that it functions as a button.... So everywhere i put a transparent button with the size of the wrapper layout and then the Talkbar reads the text and automatically the "button" part so no string res change needed.... also for the user now the whole "wrapper" (the full button size) is selected (image + text) so it is also more user friendly


a-szotyori (approved):
Great work! Thank you!

a-szotyori:
Will be QA'd in the release candidate.

PR #289, opened 2023-04-05
Pia 3926 help accessibility talkbar

llevente:
[NOTE]: PIA-3930 also in this PR

[NOTE]: Problem occured when bottom bar was activated so we used in the topbar the menu items... it seems like for menu items the title overrides the contentDescription so i changed the title. This could be an issue if more menus will be added and there will be no space for icons so the dropdown list shows with the titles.... Please consider if this can be a real issue.

### DIFF
--- a/capture-sdk/sdk/src/main/res/values-en/strings.xml
+++ b/capture-sdk/sdk/src/main/res/values-en/strings.xml
@@ -11,6 +11,7 @@
     <string name="gc_title_no_results">No results</string>
     <string name="gc_title_error">Error</string>
     <string name="gc_help">Help</string>
+    <string name="gc_help_button_content_description">Help button</string>
 
     <string name="gc_camera_error_no_permission">We need access to your camera to enable you to capture your invoices</string>
     <string name="gc_camera_error_no_permission_button_title">Give Access</string>
--- a/capture-sdk/sdk/src/main/res/values-en/strings.xml
+++ b/capture-sdk/sdk/src/main/res/values-en/strings.xml
@@ -11,8 +11,8 @@
     <string name="gc_title_no_results">No results</string>
     <string name="gc_title_error">Error</string>
     <string name="gc_help">Help</string>
-    <string name="gc_help_button_content_description">Help button</string>
-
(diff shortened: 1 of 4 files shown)


capture-sdk/sdk/src/main/res/values/strings.xml
```diff
@@ -161,4 +161,6 @@
     <string name="gc_delete_page">Seite löschen</string>
     <string name="gc_document_page">Dokumentenseite</string>
     <string name="gc_useful_tips">Nützliche Tipps</string>
+    <string name="gc_menu_item_help_button_content_description">Hilfe taste</string>
+    <string name="gc_menu_item_close_button_content_description">Schließen taste</string>
```
a-szotyori, line 165:
```suggestion
    <string name="gc_menu_item_help_button_content_description">Hilfe, Schaltfläche</string>
    <string name="gc_menu_item_close_button_content_description">Schließen, Schaltfläche</string>
```

capture-sdk/sdk/src/main/res/values-en/strings.xml
```diff
@@ -11,7 +11,8 @@
     <string name="gc_title_no_results">No results</string>
     <string name="gc_title_error">Error</string>
     <string name="gc_help">Help</string>
-
+    <string name="gc_menu_item_help_button_content_description">Help button</string>
+    <string name="gc_menu_item_close_button_content_description">Close button</string>
```
a-szotyori, line 15:
```suggestion
    <string name="gc_menu_item_help_button_content_description">Help, button</string>
    <string name="gc_menu_item_close_button_content_description">Close, button</string>
```

a-szotyori (changes requested):
Thank you for the workaround. TalkBack's behaviour is weird. In other apps it says "button" on menu items. I also tried our old UI which uses the default action bar and there it also doesn't say "button".

Well, our menu items are set to be always shown as action so the titles should never be visible until we don't have to add more menu items.

I left two comments to add a pause before TalkBack says "button" and to use the same German word for button that TalkBack uses ([Schaltfläche](https://de.wikipedia.org/wiki/Schaltfl%C3%A4che), a synonym for Taste).

PR #299, opened 2023-04-25
fix(bank-sdk): Fixed example app UI

llevente:


a-szotyori (approved):
Thanks! Can go into QA.

PR #300, opened 2023-04-27
feat(capture-sdk): Added No results for QR scan

llevente:


a-szotyori (approved):
Looks good! Can be merged directly.

PR #306, opened 2023-05-09
Docs for predictive back gesture - 1.x.x

a-szotyori:


zladzeyka (approved):
👍

zladzeyka (changes requested):
[Clarification] Could you please confirm that  `Content type: Getting started` - we are inside the integration guide?

PR #307, opened 2023-05-09
Docs for predictive back gesture - 2.x.x

a-szotyori:


bank-sdk/sdk/src/doc/source/integration.rst
```diff
@@ -1,6 +1,17 @@
 Integration
 ===========
 
+..
+  Audience: Android dev who integrates for the first time
+  Purpose: Describe what app configuration is needed, which preconditions have to be met, how to configure the SDK and how to run it.
+  Content type: Getting started
```
zladzeyka, line 7: [Clarification] Just to check it one more time `Content type: Getting started` and we are inside the Integration file, is it correct?

zladzeyka (changes requested):
Looks good, just a small clarification question 🤓

zladzeyka (approved):

PR #315, opened 2023-06-01
1.x.x - Don't delete image multi-page documents when leaving the camera screen via cancellation

a-szotyori:
It's not critical for our backend that unused documents are deleted.

By removing deletion when leaving the camera screen due to cancellation we eliminate unwanted deletions caused by an unexpected lifecylce behaviour change in Android 13 (onStart is called before onActivityResult in Android 13).

### DIFF
--- a/capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragmentImpl.java
+++ b/capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragmentImpl.java
@@ -642,13 +642,8 @@ class CameraFragmentImpl implements CameraFragmentInterface, PaymentQRCodeReader
             mImportUrisAsyncTask.cancel(true);
         }
 
-        if (!mInstanceStateSaved) {
-            if (!mProceededToMultiPageReview) {
-                clearMultiPageDocument();
-            }
-            if (!mQRCodeAnalysisCompleted) {
-                deleteUploadedQRCodeDocument();
-            }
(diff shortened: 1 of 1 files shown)


capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragmentImpl.java
```diff
@@ -661,40 +660,6 @@ private void clearMultiPageDocument() {
...
-                                    networkRequestsManager.delete(giniCaptureDocument);
-                                }
-                                return null;
-                            }
-                        });
-            }
-        }
-    }
-
     private void deleteUploadedQRCodeDocument() {
```
zladzeyka, line 663: [Clatification] @a-szotyori do we still use `deleteUploadedQRCodeDocument` somewhere?

zladzeyka: The QR code deletion is only called if the camera screen is closed before the extractions arrive.

zladzeyka (approved):

capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragmentImpl.java
a-szotyori: To be safe, I removed qr code document deletion.

zladzeyka (approved):

PR #322, opened 2023-07-12
Pia 4344 field entry point config option

a-szotyori:


capture-sdk/sdk/src/main/java/net/gini/android/capture/EntryPoint.kt
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Created by Alpár Szotyori on 10.07.23.
+ *
+ * Copyright (c) 2023 Gini GmbH.
+ */
+
+/**
+ * Use this enum to set which entry point is used for starting the SDK.
+ */
+enum class EntryPoint {
```
zladzeyka, line 12: [Clarification]: What do you think about putting the **BUTTON** at the 1st enum position?
It's going to be a default value and alphabetically is higher.

capture-sdk/sdk/src/test/assets/remslip-malformed-user-comment.jpeg
zladzeyka, line 1: [Clarification]: Can we use this document in the app? Maybe it's better to pick one without the real name or an anonymized one ;)

mahdiabolfazli (commented):
Everything seems fine. 
[Suggestion] Why not refactoring into Kotlin when touching a file? It would not take long I guess :)

a-szotyori:
@mahdiabolfazli Thanks for pointing out converting to kotlin. I refactored the tests to kotlin when I ported them from `androidTest` to `test`. I didn't refactor other Java classes to Kotlin for the following reasons:
* Reduce sources of error - the Java files I touched were mission critical.
* Keep scope small.
* Java still works fine.
* Only benefit from Kotlin is that we modernize the syntax. No plans currently to use Kotlin only features in the touched Java files.

capture-sdk/sdk/src/test/assets/remslip-malformed-user-comment.jpeg
a-szotyori: Great point. I will replace them with an anonimzed jpeg to be on the safe side.

zladzeyka (approved):
👏🏻

zladzeyka (approved):

PR #323, opened 2023-07-18
PIA-4339 Fix error screen result handling for QR codes in the camera screen

a-szotyori:
The results of the qr code error screens that were started from the camera screen were ignored. This fix makes sure the results are handled.


mahdiabolfazli (approved):
Everything seems fine to me.

PR #325, opened 2023-07-21
Pia 4412 use capital case "Bearer" instead of all caps "BEARER"

a-szotyori:
The change is simple and I used the occasion to improve the bearer and basic related headers code while also adding tests to ensure that capital case "Bearer" and "Basic" is used in authorisation headers.


mahdiabolfazli (approved):
I ran the tests and reviewed the code, everything seems great. Good job 🥇

mahdiabolfazli (approved):

PR #326, opened 2023-07-21
Pia 4412 2.x.x use capital case "Bearer" instead of all caps "BEARER"

a-szotyori:
Updates the 2.x.x versions of Bank APi Library, Health API Library and Core API Library.

The change is simple and I used the occasion to improve the bearer and basic related headers code while also adding tests to ensure that capital case "Bearer" and "Basic" is used in authorisation headers.

### DIFF
--- a/core-api-library/library/src/main/java/net/gini/android/core/api/authorization/UserRemoteSource.kt
+++ b/core-api-library/library/src/main/java/net/gini/android/core/api/authorization/UserRemoteSource.kt
@@ -7,6 +7,7 @@ import net.gini.android.core.api.authorization.apimodels.SessionTokenInfo
 import net.gini.android.core.api.authorization.apimodels.UserRequestModel
 import net.gini.android.core.api.authorization.apimodels.UserResponseModel
 import net.gini.android.core.api.requests.ApiException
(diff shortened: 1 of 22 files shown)


core-api-library/library/src/main/java/net/gini/android/core/api/authorization/UserRemoteSource.kt
```diff
@@ -61,11 +62,11 @@ internal class UserRemoteSource(
     private fun basicHeaderMap(): Map<String, String> {
         val encoded = Base64.encodeToString("${clientId}:${clientSecret}".toByteArray(), Base64.NO_WRAP)
         return mapOf("Accept" to "application/json",
```
mahdiabolfazli, line 64: [Suggestion] what about adding another class to `BasicAuthorizatonHeader` for the `mapOf("Accept" to "application/json"` ?

a-szotyori: I agree, thanks for spotting this.

mahdiabolfazli (approved):
I ran the tests and reviewed the code, everything seems great. Good job 🥇

mahdiabolfazli (approved):

PR #332, opened 2023-08-01
PIA-4435-update-reference-docs

a-szotyori:
* Fix the displayed module name on the generated dokka webpages.
* Remove higher level information that was outdated and/or duplicated information from our main documentation.


abolfazlimahdi (approved):
[Suggestion] I noticed that in `bank-api-library` and `bank-sdk`, you used the import for `DokkaCollectorTask` class. However, for other modules, you have used the full package name for the class. I think it would be better to be consistent in all of them. Otherwise, everything else seems fine. :)

a-szotyori:
Thank you! I will fix the inconsistency and use imports.

PR #334, opened 2023-08-22
feat(capture-sdk): Remove storage permissions

abolfazlimahdi:
- Remove READ_MEDIA_IMAGES and READ_EXTERNAL_STORAGE permissions
- Remove classes for storage permissions

PIA-4282


capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/fileimport/FileChooserActivity.java
```diff
@@ -318,153 +281,27 @@ public void onItemSelected(@NonNull final ProvidersAppItem item) {
                         // Store the selected item, it is needed for requesting permission
                         // after showing the rationale
                         mSelectedAppItem = item;
```
a-szotyori, line 283: [Request] It looks like the `mSelectedAppItem` is not needed anymore. It was only needed to know which app to launch after the permission request. I think it is safe to remove it. Can you please double check and remove it, if possible? Thanks!

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/fileimport/FileChooserActivity.java
```diff
@@ -318,153 +281,27 @@ public void onItemSelected(@NonNull final ProvidersAppItem item) {
...
-                .setMessage(R.string.gc_storage_permission_rationale)
-                .setPositiveButton(R.string.gc_storage_permission_rationale_positive_button)
-                .setNegativeButton(R.string.gc_storage_permission_rationale_negative_button)
-                .setDialogId(PERMISSION_RATIONALE_DIALOG)
-                .disableCancelOnTouchOutside()
-                .create();
-        showPermissionDialog(dialogFragment);
-    }
 
     private void showAppDetailsSettingsScreen() {
```
a-szotyori, line 301: [Request] This method is not used anymore, you can remove it.

a-szotyori (changes requested):
Great work! I found two more things which could be removed.

a-szotyori (approved):
Nice! This PR can go into QA, but we also need this change for 1.x and 2.x first.

PR #337, opened 2023-08-23
fix(core-api-library): Fix multiple user creation

abolfazlimahdi:
- created a test to fail in case it happens again

PIA-4447


core-api-library/library/src/main/java/net/gini/android/core/api/DocumentRepository.kt
```diff
@@ -29,6 +31,13 @@ abstract class DocumentRepository<E: ExtractionsContainer>(
     private val giniApiType: GiniApiType
 ) {
 
+    /*
+    We need mutex lock because otherwise when the user upload multiple documents at first use of the app, the app
+    creates multiple users (equal to number of documents) and we will get a server error (the document does not belong to the user)!
+    we are using this Mutex in the withAccessToken() function so that it affect the whole application
```
a-szotyori, line 37: [Suggestion]
```suggestion
    We are using a Mutex to prevent coroutines from retrieving access tokens in parallel. This way even when multiple uploads are started only one user is created.
```

core-api-library/library/src/main/java/net/gini/android/core/api/DocumentRepository.kt
```diff
@@ -29,6 +31,13 @@ abstract class DocumentRepository<E: ExtractionsContainer>(
...
+     */
+    val mutex = Mutex()
```
a-szotyori, line 39: [Suggestion]
```suggestion
    val accessTokenMutex = Mutex()
```

a-szotyori (commented):
I have two suggestions.

a-szotyori (approved):

PR #338, opened 2023-08-23
fix(capture-sdk): Accept EPC069-12 QR codes with CR (\r) line delimiters

a-szotyori:
Even though according to the standard only LF (\n) or CRLF (\r\n) line delimiters are allowed some invoice issuers use CR line delimiters.

PIA-4468


abolfazlimahdi:
Everything is fine. There is another issue with QRCodes that I think should be handled in another ticket. The data extracted from [this QRCode](https://wiki.i.gini.net/download/thumbnails/131367276/Screenshot%202023-08-23%20at%2015.14.00.png?version=1&modificationDate=1692796450445&api=v2) is incorrect.

abolfazlimahdi (approved):
Everything is fine. There is another issue with QRCodes that I think should be handled in another ticket. The data extracted from [this QRCode](https://wiki.i.gini.net/download/thumbnails/131367276/Screenshot%202023-08-23%20at%2015.14.00.png?version=1&modificationDate=1692796450445&api=v2) is incorrect.

PR #340, opened 2023-08-24
Pia 4473 fix drawable overriding

a-szotyori:


capture-sdk/sdk/build.gradle.kts
```diff
@@ -31,6 +31,12 @@ android {
 
         buildConfigField("String", "VERSION_NAME", "\"$version\"")
         buildConfigField("String", "VERSION_CODE", "\"${properties["versionCode"]}\"")
+
```
abolfazlimahdi, line 34: [Suggestion] I think this comment is somehow ambiguous. Maybe I got it wrong but I am proposing a suggestion for it. Use it if you feel it is better.

// Prevent creating PNGs for vector drawables. Without this, PNGs would override vector 
        // drawables which means we should add PNGs to both 'drawable-anydpi' and 'drawable-anydpi-v24'.
        // Our min SDK level allows us to use the runtime support for
        //  vector drawables so it's safe and we also benefit from reduction in the apk size.

abolfazlimahdi (approved):
Just added a suggestion for the comment. I also tested the behavior and it was working as expected. Great 💯

capture-sdk/sdk/build.gradle.kts
a-szotyori: Oh, I see my phrasing caused misunderstanding. Unfortunately your rephrasing changes the meaning. PNGs should override vector drawables so we need `useSupportLibrary` to enable overriding. Also we need `useSupportLibrary` to avoid having to add vector drawables twice (in `drawable-anydpi` and `drawable-anydpi-v24`).

Is this version clearer?

// Setting 'vectorDrawables.useSupportLibrary = true' prevents creating PNGs for vector drawables.
// This is needed to allow overriding of drawables either with other vector drawables added to 'drawable-anydpi'
// or with PNGs.
// If 'vectorDrawables.useSupportLibrary' is not set to 'true' then PNGs won't override vector drawables and
// overriding of vector drawables with other vector drawables only works on API Level 24 and higher if they
// are also added to 'drawable-anydpi-v24' in addition to 'drawable-anydpi'.
// Our min SDK level allows only Android versions which have runtime support for vector drawables
// so it's safe to prevent creating PNGs from vector drawables and we also reduce the apk size.

a-szotyori:
@abolfazlimahdi I added PNGs for those vector drawables that were not working correctly or caused a crash on API Levels 21 to 23. I did this for Capture SDK, Bank SDK and Health SDK.

PR #343, opened 2023-08-28
fix(capture-sdk): Add background color resource for file provider

a-szotyori:
PIA-4477


abolfazlimahdi (approved):
Everything works as expected 💯

PR #344, opened 2023-08-28
fix(capture-sdk): Replace usage of Java 8 stream mapping with Kotlin collections' `map()` function

a-szotyori:
PIA-4472


abolfazlimahdi (approved):
Everything looks nice :-)

PR #355, opened 2023-09-25
Pia 4547 deprecate requirements check

a-szotyori:
The majority of Android devices meet the Capture/Bank SDK's requirements. Also sometimes the requirements checking fails due issues in the Android camera apis (e.g., [PIA-2482](https://ginis.atlassian.net/browse/PIA-2482)).

By deprecating (and later removing) the requirements checking we will allow a small number of additional old devices to use the SDK. The extraction quality might be lower due to the smaller camera resolution (less than 8MP), but we won't block users anymore and also will prevent wrongly blocking users with adequate devices.

### DIFF
--- a/bank-sdk/example-app/src/main/java/net/gini/android/bank/sdk/exampleapp/ui/MainActivity.kt
+++ b/bank-sdk/example-app/src/main/java/net/gini/android/bank/sdk/exampleapp/ui/MainActivity.kt
@@ -142,11 +142,6 @@ class MainActivity : AppCompatActivity() {
     }
 
     private fun startGiniCaptureSdk(intent: Intent? = null) {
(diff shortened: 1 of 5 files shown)


abolfazlimahdi (approved):
[Comment] Are you sure we want to deprecate the whole requirements? For example, there is a camera requirement (if the device has camera) or memory requirement as well. Just noticed that these may have been overlooked :-) otherwise, everything else looks good.

a-szotyori:
@abolfazlimahdi Yes, all requirements can be put to rest. Due to advancements in smartphone hardware the likelyhood of a device not having an 8MP or better camera (with flash and autofocus) or low memory is very low by now in our region (western and central europe).

PR #356, opened 2023-10-09
Pia 4510 UI for iban detection on device

abolfazlimahdi:


capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/iban/IBANRecognizerFilter.kt
```diff
@@ -29,12 +29,22 @@ class IBANRecognizerFilter @JvmOverloads constructor(
                     if (prevIBANs.isNotEmpty() && ibans.isNotEmpty()) {
                         delay(1000)
                     }
-                    listener.onIBANsReceived(ibans)
+                    listener.onIBANsReceived(ibans.setIBANFormatting())
```
a-szotyori, line 32: [Request] We also need a test for this 😉

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/iban/IBANRecognizerFilter.kt
```diff
@@ -29,12 +29,22 @@ class IBANRecognizerFilter @JvmOverloads constructor(
...
                     prevIBANs = ibans
                 }
         }
     }
 
+    private fun List<String>.setIBANFormatting(): List<String> {
```
a-szotyori, line 38: [Suggestion] Private extension functions which are used only in one place seem overkill when a plain old private function can do the trick, for ex. `applyIBANFormatting(ibans)`.

As per Kotlin [documentation](https://kotlinlang.org/docs/extensions.html) extension functions are intended to add functionality without having to inherit. In this case we could avoid using extensions and just use plain old simpler code. 🤓

a-szotyori (changes requested):
Looks good, only the test is missing and the extension function could be simplified to a plain old private function.

PR #368, opened 2023-11-08
fix(bank-sdk): Fix example app launching the Bank SDK multiple times when using "open with"

a-szotyori:


abolfazlimahdi (approved):
Perfect 🥇

PR #369, opened 2023-11-14
feat(capture-sdk): Disable flash by default

abolfazlimahdi:
BSDK-95

### DIFF
--- a/bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/capture/Configuration.kt
+++ b/bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/capture/Configuration.kt
@@ -87,9 +87,9 @@ data class CaptureConfiguration(
     val flashButtonEnabled: Boolean = false,
 
     /**
-     * Set whether the camera flash is on or off by default.
+     * Set whether the camera flash should be on or off when the SDK starts. The flash is off by default.
      */
-    val flashOnByDefault: Boolean = true,
+    val flashOnByDefault: Boolean = false,
 
     /**
      * Enable/disable the return assistant feature.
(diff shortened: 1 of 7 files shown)


bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/capture/Configuration.kt
```diff
@@ -89,7 +89,7 @@ data class CaptureConfiguration(
     /**
      * Set whether the camera flash is on or off by default.
```
a-szotyori, line 90: [Suggestion] What do you think of repharsing it to:
```suggestion
     * Set whether the camera flash should be on or off when the SDK starts. The flash is off by default.
```

bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/capture/Configuration.kt
```diff
@@ -89,7 +89,7 @@ data class CaptureConfiguration(
...
      */
-    val flashOnByDefault: Boolean = true,
+    val flashOnByDefault: Boolean = false,
```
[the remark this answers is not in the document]
abolfazlimahdi: Done! Thanks for noticing :)

bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/capture/Configuration.kt
abolfazlimahdi: Done! Also changed for capture docs :-)

a-szotyori (approved):
Nice!

a-szotyori (approved):
👏

PR #373, opened 2023-11-28
fix(bank-sdk): Fix example app bugs

abolfazlimahdi:
- fix app crash when switching to dark mode in configuration screen
- fix close button color in dark mode


a-szotyori (approved):
Thanks!

PR #377, opened 2024-02-01
Pp 226 prevent screenshots

a-szotyori:
- Adds a new configuration option to Capture SDK and Bank SDK to allow or disallow screenshots. By default screenshots are allowed.
- Also adds this new configuration option to the example app's configuration screen.

The new configuration option can be used as follows:
* Capture SDK: 
   ```
    GiniCapture.newInstance(context)
        .setAllowScreenshots(false) // or true
        ...
        .build()
   ```
* Bank SDK: 
   ```
        GiniBank.setCaptureConfiguration(context, CaptureConfiguration(
            allowScreenshots = false, // or true
            ...
        ))
   ```

Note: builds upon the host fragments branch which we plan to be our upcoming version. In the extreme case that our release of that is delayed we can easily port this to work with latest version we have in `main`.


abolfazlimahdi (approved):
Nice job. Great 🥇

PR #379, opened 2024-02-08
Ipc 108 payment component with pay invoice button

a-szotyori:
Health SDK changes:
- Created view for the Payment Component (always shows Sparkasse icon and name for demoing purposes)
- Updated theme to be based on Material 3 with dark and light mode support (based on how it is done in Capture SDK)
- Added color palette and typography (based on how it is done in Capture SDK)

Example app changes:
- Updated main screen and theme to be more similar to the Bank SDK example app
- Updated icon to use the same yellow logo we use on iOS
- Checks if invoice is payable during upload and analysis (after extractions are available) and store it with the document
- Showing the Payment Component view for payable invoices


abolfazlimahdi (approved):
Nice job 🥇

PR #380, opened 2024-02-14
Ipc 111 show first payment provider or placeholder text

a-szotyori:
Adds the `PaymentComponentsController` which loads the payment providers from the Health API and links them to installed payment provider apps.

The `PaymentComponentsController` instance needs to be passed to each `PaymentComponentView` so that each view can observe the list of payment provider apps and update their subviews accordingly.


abolfazlimahdi (approved):
everything is fine. just two things to consider:

1. I saw three TODO s in the code, do you keep track of them somewhere else so they are not ignored in the future (maybe in the related tickets in Jira)?
2. As we talked yesterday, maybe renaming the PaymentComponentsController is a good idea (just a reminder, not that it should be done)

a-szotyori:
@abolfazlimahdi Thank you for the review! I renamed `PaymentComponentsController` to `PaymentComponent` and added the relevant Jira story links to the todos.

PR #389, opened 2024-03-04
Pp 312 pp 313 fix iban detector not enabled and fix focus indicator position

a-szotyori:
Fixes the following bugs:
1. IBAN detection not working if QR code scanning is disabled: https://ginis.atlassian.net/browse/PP-312
2. Wrong focus indicator position: https://ginis.atlassian.net/browse/PP-313


abolfazlimahdi (approved):
Perfect. Thanks 👍

PR #391, opened 2024-03-06
IPC-168 - Remove requirements check for installed bank apps

danicretu:
- Remove requirement check for at least 1 bank app to be installed
- Remove corresponding tests
- Remove requirement references from documentation


abolfazlimahdi (approved):
Nice job. Thanks 🥇

PR #393, opened 2024-03-15
Ipc 190 unit tests more information

danicretu:


health-sdk/sdk/src/test/java/net/gini/android/health/sdk/moreinformation/MoreInformationTest.kt
```diff
@@ -0,0 +1,102 @@
...
+
+    @Test
+    fun `displays all FAQ questions`() {
+        // When
+        launchFragmentInContainer {
+            MoreInformationFragment.newInstance(paymentComponent)
+        }
+
+        // Then
+        onView(withId(R.id.ghs_faq_list)).check { view, _ -> assertThat ((view as ExpandableListView).adapter!!.count).isEqualTo(6) }
```
a-szotyori, line 100: [Suggestion] This test will fail if we add or remove questions. What do you think of making the list of FAQs internally accessible (and adding the `@VisibleForTesting` to make it explicit why it's not private) and then use the list's length in the assertion here?

Alternatively you could pass the faq list as an argument to the fragment's `newInstance()`, but that needs a bit bigger refactoring. No need to do that now because we might refactor this screen in the future to fetch the questions from our API. Until then we keep the questions hardcoded in the fragment.

a-szotyori (commented):
Looks great! I left a suggestion for one of the tests.

health-sdk/sdk/src/test/java/net/gini/android/health/sdk/moreinformation/MoreInformationTest.kt
danicretu: Thank you for the suggestions - I went with the first option since there is a possibility we'll need to refactor to get the FAQ's from the API.

a-szotyori (approved):
Awesome, thank you! Btw this needs no QA and you can merge it and move the ticket to done. 🎉

PR #395, opened 2024-03-15
Ipc 186 unit tests payment component

danicretu:


health-sdk/sdk/src/test/java/net/gini/android/health/sdk/paymentComponent/PaymentComponentViewTest.kt
```diff
@@ -0,0 +1,168 @@
...
+class PaymentComponentViewTest {
+    private var activity : Activity? = null
+    private var context: Context? = null
+    private var scenario: ActivityScenario<Activity>? = null
+    private lateinit var paymentComponent: PaymentComponent
+    private lateinit var paymentComponentListener: PaymentComponent.Listener
+
+    @Before
+    fun setUp() {
+        activity = Robolectric.buildActivity(Activity::class.java).create().get()
```
a-szotyori, line 37: [Suggestion] `activity` is not used and I checked quickly that it also works if we just call `Robolectric.buildActivity...` without assigning it to anything:
```suggestion
        Robolectric.buildActivity(Activity::class.java).create().get()
```

Maybe move it down to line 44 and leave comment that it's needed for the ActivityScenario because it's not obvious why we need it.

a-szotyori (changes requested):
Great work! I think we need three more tests and then it's done.

health-sdk/sdk/src/main/java/net/gini/android/health/sdk/util/extensions/Context.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import android.content.Context
+import android.graphics.Bitmap
+import android.graphics.BitmapFactory
+import android.graphics.drawable.BitmapDrawable
+import android.util.TypedValue
+import androidx.annotation.VisibleForTesting
+import net.gini.android.health.sdk.paymentprovider.PaymentProviderApp
+
+// In a future refactoring we can split extensions into files according to what component they extend
```
a-szotyori, line 11: [Comment] Yes, thank you for adding this comment! I wholeheartedly agree! 👍 👍 👍

health-sdk/sdk/src/test/java/net/gini/android/health/sdk/paymentComponent/PaymentComponentTest.kt
```diff
@@ -100,17 +104,50 @@ class PaymentComponentTest {
...
+        val resources: Resources = spyk(context!!.resources)
+        val packageManager: PackageManager = mockk(relaxed = true)
+
+        mockkStatic(Context::generateBitmapDrawableIcon)
+        mockkStatic(PackageManager::getPaymentProviderApps)
+        every { privateContext.applicationContext } returns context
+        every { privateContext.packageManager } returns packageManager
+        every { privateContext.resources } returns resources
+        every { privateContext.generateBitmapDrawableIcon(any(), any()) } returns null
+        every { packageManager.getPaymentProviderApps(paymentProviderList, privateContext) } returns paymentProviderAppsList
```
a-szotyori, line 128: [Comment] Thank you for this elegant solution! 👏

a-szotyori (approved):
Great job! Thank you that you persevered and find such elegant solution to the context mocking problem! 👏

PR #398, opened 2024-03-19
Ipc 211 small UI issues

danicretu:


a-szotyori (approved):
Thank you for the fixes!

PR #400, opened 2024-03-21
Pp 330 pp332 fix return assistant typos and disable return reasons

abolfazlimahdi:


bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/GiniBank.kt
```diff
@@ -106,7 +106,7 @@ object GiniBank {
     /**
      * Shows the return reasons dialog in the return assistant, if enabled.
      */
```
a-szotyori, line 108: [Request] Can you please add a sentence here that it's disabled by default? Thanks!

bank-sdk/sdk/src/test/java/net/gini/android/bank/sdk/GiniBankTest.kt
```diff
@@ -79,6 +79,6 @@ class GiniBankTest {
 
     @Test
     fun `return reasons dialog is enabled by default`() {
```
a-szotyori, line 81: [Request] Please also update the test name. 🙏

a-szotyori (changes requested):
Looks great, I have just two small requests.

a-szotyori (approved):

PR #402, opened 2024-03-21
fix(health-sdk): Fixed tests failing after running tests on `MoreInfo…

danicretu:
…rmation` related classes

### DIFF
--- a/health-sdk/sdk/src/main/java/net/gini/android/health/sdk/moreinformation/MoreInformationViewModel.kt
+++ b/health-sdk/sdk/src/main/java/net/gini/android/health/sdk/moreinformation/MoreInformationViewModel.kt
@@ -23,25 +23,34 @@ internal class MoreInformationViewModel(private val paymentComponent: PaymentCom
         _paymentProviderAppsListFlow.asStateFlow()
 
     fun start() {
-        if (paymentComponent == null) {
-            LOG.warn("Cannot show payment provider apps: PaymentComponent must be set before showing the BankSelectionBottomSheet")
-            return
-        }
-        LOG.debug("Collecting payment provider apps state and selected payment provider app from PaymentComponent")
-        paymentComponent.paymentProviderAppsFlow.onEach { paymentProviderAppsState ->
-            when (paymentProviderAppsState) {
-                is PaymentProviderAppsState.Error -> { processError(paymentProviderAppsState.throwable) }
-                PaymentProviderAppsState.Loading -> {}
-                is PaymentProviderAppsState.Success -> {
-                    if (paymentProviderAppsState.paymentProviderApps.isEmpty()) {
-                        LOG.debug("No payment provider apps received")
-                    } else {
-                        LOG.debug("Received {} payment provider apps", paymentProviderAppsState.paymentProviderApps.size)
+        viewModelScope.launch {
+            if (paymentComponent == null) {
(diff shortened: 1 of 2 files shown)


health-sdk/sdk/src/main/java/net/gini/android/health/sdk/moreinformation/MoreInformationViewModel.kt
```diff
@@ -21,27 +23,25 @@ internal class MoreInformationViewModel(private val paymentComponent: PaymentCom
...
+                    if (paymentProviderAppsState.paymentProviderApps.isEmpty()) {
+                        LOG.debug("No payment provider apps received")
+                    } else {
+                        LOG.debug("Received {} payment provider apps", paymentProviderAppsState.paymentProviderApps.size)
                     }
+                    _paymentProviderAppsListFlow.value = PaymentProviderAppsListState.Success(paymentProviderAppsState.paymentProviderApps)
                 }
             }
-        }
+        }.launchIn(viewModelScope)
```
a-szotyori, line 44: [Clarification] Can you help me understand why this change is needed? Thank you!

a-szotyori (commented):
I just wish to understand the reason for this change.

a-szotyori (approved):

health-sdk/sdk/src/main/java/net/gini/android/health/sdk/moreinformation/MoreInformationViewModel.kt
a-szotyori: [Request] I investigated a bit and found that the test failure was caused by mocking. Actually the lack of providing a value for the `paymentProviderAppsFlow` on the mocked `paymentComponent` in the `MoreInformationTest`.

According to the documentation `launchIn(scope)` is a shorthand for `scope.launch { flow.collect { } }` so this shouldn't have fixed the test failure. Maybe the `onEach()` call prevented the `kotlin.KotlinNothingValueException` exception.

Anyway simply adding a value for the flow in the `setup()` function of `MoreInformationTest.kt` solved the test failure without having to change the `MoreInformationViewModel`:
```
(...)
class MoreInformationTest {
(...)
@Before
fun setup() {
    paymentComponent = mockk(relaxed = true)
    // Provide a default empty flow to prevent kotlin.KotlinNothingValueException exceptions
    every { paymentComponent!!.paymentProviderAppsFlow } returns MutableStateFlow<PaymentProviderAppsState>(mockk()).asStateFlow()
}
(...)
```

a-szotyori (changes requested):
I did a bit of investigation and found that the test failure can be solved without changing the `MoreInformationViewModel`.

health-sdk/sdk/src/main/java/net/gini/android/health/sdk/moreinformation/MoreInformationViewModel.kt
a-szotyori: Oh, I just saw your comment. Now I know why `onEach` fixed. So if `onEach` wraps `paymentComponent.paymentProviderFlow` than we can do that explicitly in the test ourselves by providing an empty flow for `paymentComponent.paymentProviderFlow` by default in the test setup method.

a-szotyori:
@danicretu Sorry, Daniela, I again applied changes to your PR. I want to create a release candidate for Syeda today. For that I need to merge this fix first.

a-szotyori (approved):

PR #403, opened 2024-03-22
fix(health-sdk): Keep same pager height on configuration change in `R…

danicretu:
…eviewFragment`

IPC-201


health-sdk/sdk/src/main/java/net/gini/android/health/sdk/review/ReviewFragment.kt
```diff
@@ -214,7 +215,6 @@ class ReviewFragment private constructor(
                     indicator.isVisible = pages.size > 1
                     pager.isUserInputEnabled = pages.size > 1
                 })
-                removePagerConstraint()
```
danicretu, line 217: [Note] Removed this because it was already being called before in `onViewCreated`.

a-szotyori (approved):
Thank you for fixing this!

PR #409, opened 2024-03-26
Pp 302 allow finishing of open with activity

danicretu:
- Exposed document processing method 
- Added methods to forward document to activity / fragment

- Added `SplashActivity` to forward document to `MainActivity`, then finish
- Added `SplashSingleActivity` to forward document to `CaptureFlowHostActivity`, then finish
- Extended functionality to support loading a document directly in a fragment to exemplify how it would work in single activity applications.


bank-sdk/example-app/src/main/java/net/gini/android/bank/sdk/exampleapp/ui/CaptureFlowHostActivity.kt
```diff
@@ -21,7 +23,13 @@ class CaptureFlowHostActivity : AppCompatActivity() {
...
             val openWithIntent = IntentCompat.getParcelableExtra(intent, EXTRA_IN_OPEN_WITH_INTENT, Intent::class.java)
             if (openWithIntent != null) {
-                startBankSDKForOpenWith(openWithIntent)
+                if (!intent.hasExtra(ExampleUtil.DOCUMENT)) {
+                    startBankSDKForOpenWith(openWithIntent)
+                } else {
+                    intent.getParcelableExtra(ExampleUtil.DOCUMENT, Document::class.java)?.let {
+                        startBankSDKForDocument(it)
+                    }
+                }
```
a-szotyori, line 32: [Suggestion] Can we separate the "open with" intent case from the "open with" document one?

I would suggest to move the "has document extra" check out of the "has open with intent" check. Also to use `IntentCompat.getParcelableExtra()` because it's backwards compatible:
```
if (openWithIntent != null) {
    startBankSDKForOpenWith(openWithIntent)
} else if (intent.hasExtra(ExampleUtil.DOCUMENT) {
    IntentCompat.getParcelableExtra(intent, ExampleUtil.DOCUMENT, Document::class.java)?.let {
        startBankSDKForDocument(it)
    }
}

File: bank-sdk/example-app/src/main/java/net/gini/android/bank/sdk/exampleapp/ui/CaptureFlowHostActivity.kt [hunk 2]
```diff
@@ -37,10 +45,18 @@ class CaptureFlowHostActivity : AppCompatActivity() {
             .startCaptureSDKForIntent(openWithIntent)
     }
 
+    private fun startBankSDKForDocument(document: Document) {
+        findViewById<FragmentContainerView>(R.id.fragment_host).getFragment<ClientBankSDKFragment>()
+            .startBankSDKForDocument(document)
+    }
+
     companion object {
         fun newIntent(context: Context, openWithIntent: Intent? = null) =
```
a-szotyori, line 54: [Suggestion] What do you think of creating a second `newIntent()` method which can be used for the document:
```
fun newIntent(context: Context, openWithDocument: Document)
```

It looks like that have three distinct entry points for this activity:
1. launch without extras
2. launch with "open with" Intent
3. launch with "open with" Document

We should then reflect this in the `CaptureFlowHostActivity`'s public API, no?

bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/GiniBank.kt
```diff
@@ -406,6 +406,79 @@ object GiniBank {
...
+
+                override fun onCancelled() {
+                    callback(CreateDocumentFromImportedFileResult.Cancelled)
+                }
+            }
+        )
+    }
+
+    /**
+     * Starts review flow for a processed document.
```
a-szotyori, line 462: [Request]
```suggestion
     * Starts capture flow for a document. This method should be used with documents created by [GiniBank.createDocumentForImportedFiles] when a pdf or image was shared from another app.
```

bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/GiniBank.kt
```diff
@@ -406,6 +406,79 @@ object GiniBank {
...
+     * @param document The document to be forwarded by the result launcher.
+     */
+    fun startCaptureFlowForDocument(            // maybe we should name it `startReviewFlowForDocument` ? seeing as there is no capture flow per se
+        resultLauncher: ActivityResultLauncher<CaptureImportInput>, document: Document
+    ) {
+        resultLauncher.launch(CaptureImportInput.Forward(document))
+    }
+
+    /**
+     *  Creates a [CaptureFlowFragment] with a processed document.
```
a-szotyori, line 474: [Request]
```suggestion
     *  Creates a [CaptureFlowFragment] with a document. This method should be used with documents created by [GiniBank.createDocumentForImportedFiles] when a pdf or image was shared from another app.
```

bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/GiniBank.kt
```diff
@@ -406,6 +406,79 @@ object GiniBank {
...
+     *
+     *  @param document The document with which the fragment will be created.
+     */
+    fun createCaptureFlowFragmentForDocument(           //maybe we should name this one `createReviewFlowFragmentForDocument` ? same comment as above
```
a-szotyori, line 478: [Request] The reason why we should keep calling it "capture flow" is to keep it simple: the capture flow does everything document capturing, handling and processing related. If the capture flow is started with a document or intent, then it skips the camera screen. If the capture flow is started without a document or intent, then it starts with the camera screen.

bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/GiniBank.kt
```diff
@@ -406,6 +406,79 @@ object GiniBank {
...
+
+        /**
+         * Document processing was cancelled.
+         */
+        object Cancelled: CreateDocumentFromImportedFileResult()
+    }
+
+    /**
+     *
+     *  API for creating documents base on files imported from another app.
```
a-szotyori, line 431: [Request]
```suggestion
     *  Create a document based on a pdf or image files imported from another app.
```

bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/GiniBank.kt
```diff
@@ -406,6 +406,79 @@ object GiniBank {
         context.startActivity(resolvedPayment.getBusinessIntent())
     }
 
+    /**
+     * Wrapper for the processing result of a document
```
a-szotyori, line 410: [Request]
```suggestion
     * Wrapper for the processing result of a document.
```

bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/GiniBank.kt
```diff
@@ -406,6 +406,79 @@ object GiniBank {
...
+    }
+
+    /**
+     *  Creates a [CaptureFlowFragment] with a processed document.
+     *
+     *  @param document The document with which the fragment will be created.
+     */
+    fun createCaptureFlowFragmentForDocument(           //maybe we should name this one `createReviewFlowFragmentForDocument` ? same comment as above
+        document: Document,
+    ): CaptureFlowFragment = Internal.createCaptureFlowFragmentForOpenWithDocument(document)
```
a-szotyori, line 480: [Suggestion] You could simply make `createCaptureFlowFragmentForOpenWithDocument` public (also move it out of `Internal`) and rename it to `createCaptureFlowFragmentForDocument`, no? We can then avoid having a "no-op wrapper" method around it.

bank-sdk/example-app/src/main/java/net/gini/android/bank/sdk/exampleapp/core/ExampleUtil.kt
```diff
@@ -6,6 +6,8 @@ import net.gini.android.capture.network.model.GiniCaptureSpecificExtraction
 import net.gini.android.core.api.models.SpecificExtraction
 
 object ExampleUtil {
+    const val DOCUMENT = "document"
```
a-szotyori, line 9: [Suggestion] I prefer to prefix keys for input extras with `EXTRA_IN_...` to distinguish them from keys for output extras.

[Suggestion] I also prefer to scope keys for extras to their activity even if the same type of extra is used by different activities. So in our case I would suggest to add an `EXTRA_IN_OPEN_WITH_DOCUMENT` to `MainActivity` and another one to `CaptureFlowHostActivity`. I know it's duplication, but I prefer this style because by listing all extra keys in the activity's file it is more obvious which input/output extras an activity supports.

a-szotyori (changes requested):
Great work and thank you for adding the single activity splash screen! I left some suggestions and requests for some minor changes or code "simplification".

bank-sdk/example-app/src/main/java/net/gini/android/bank/sdk/exampleapp/core/ExampleUtil.kt
danicretu: Thank you for the suggestion - I applied it, but called the one in `MainActivity` differently since that one was required to be public, and was conflicting with another one.

bank-sdk/sdk/src/main/java/net/gini/android/bank/sdk/GiniBank.kt
danicretu: Yes, thank you for this. I was cautious to make a change and modify a functionality I am not aware of.

bank-sdk/example-app/src/main/java/net/gini/android/bank/sdk/exampleapp/core/ExampleUtil.kt
a-szotyori: Oh, right global constant names have to be unique. Then what about moving the extras to the respective companion objects of `MainActivity` and `CaptureFlowHostActivity`? Then uniqueness is provided by the class name + constant name, e.g.:
```
class MainActivity {
    companion object {
        const val EXTRA_IN_OPEN_WITH_DOCUMENT = "EXTRA_IN_OPEN_WITH_DOCUMENT"
    }
}
```
This would also allow code completion to find the extra when we write `MainActivity.`:
Screenshot 2024-03-27 at 18 08 20

danicretu: Done, I've moved both constants in `CaptureFlowHostActivity`, for consistency's sake.

bank-sdk/example-app/src/main/java/net/gini/android/bank/sdk/exampleapp/ui/CaptureFlowHostActivity.kt
```diff
@@ -20,8 +19,13 @@ class CaptureFlowHostActivity : AppCompatActivity() {
...
             val openWithIntent = IntentCompat.getParcelableExtra(intent, EXTRA_IN_OPEN_WITH_INTENT, Intent::class.java)
-            if (openWithIntent != null) {
-                startBankSDKForOpenWith(openWithIntent)
+            val openDocument = IntentCompat.getParcelableExtra(intent, EXTRA_IN_OPEN_WITH_DOCUMENT, Document::class.java)
+            if (openDocument != null) {
+                startBankSDKForDocument(openDocument)
+            } else if (openWithIntent != null) {
+                intent.getParcelableExtra(EXTRA_IN_OPEN_WITH_DOCUMENT, Document::class.java)?.let {
+                    startBankSDKForDocument(it)
+                }
```
a-szotyori, line 28: @danicretu I think here we should directly use the `openWithIntent`:
```suggestion
                startBankSDKForOpenWith(openWithIntent)
```

danicretu: Fixed, thanks for catching this.

a-szotyori (approved):
Thank you! Please upload the qa release build from the [Check Bank SDK](https://github.com/gini/gini-mobile-android/actions/runs/8524053856) workflow to the Android Bank SDK example app's [Internal group](https://appcenter.ms/orgs/Gini-Team-Organization/apps/Screen-API-Example-Gini-Pay-Bank-SDK/distribute/distribution-groups/Internal/releases) in App Center. 🙏

abolfazlimahdi (approved):
Good job. Thanks :-)

PR #410, opened 2024-03-26
Ipc 219 payment review screen UI customization guide

a-szotyori:
- Adds customization guide for the Payment Review Screen.
- Small adjustments of resources and UI.

You can download the rendered integration guide from the Checks tab -> Build docs for Health SDK -> Click on the "health-sdk-documentation" artifact. This will download the `health-sdk-documentation.zip`. 

Extract the zip, and open `html/index.html` to view the integration guide.


danicretu (commented):
[Clarification] Seeing as we're including the hint string resources, do we need to also include the string resources for the validation errors? (e.g. `ghs_error_input_invalid_iban`)

zladzeyka (approved):

danicretu (approved):

health-sdk/sdk/src/main/res/color-night/ghs_bottom_sheet_scrim.xml
abolfazlimahdi, line 1: [SUGGESTION] Can we have this color in the colors.xml file and reference it here? Also for the light mode. Thanks

abolfazlimahdi (commented):
Everything looks great. Just added one comment. This branch is already merged. So, we can refactor it (if applicable) later. Thanks 👍

health-sdk/sdk/src/main/res/color-night/ghs_bottom_sheet_scrim.xml
a-szotyori: The reason why I hardcoded the colors here is that this is a color resource itself (used via `R.color.ghs_bottom_sheet_scrim` or `@color/ghs_bottom_sheet_scrim`) and we don't want clients to customise it via the color palette.

It is more consistent though to use a color reference so I'll extract the colors (but won't add them to the color palette because we don't want clients to modify it).

PR #413, opened 2024-04-04
IPC-178- Accessibility settings for Payment Review

danicretu:
- Added more explanatory content description for `Amount` validation error
- Removed constraint of `Failed to get payment details` snackbar when font scale is very large (previously `snackbar` was not showing anything when font size was very large and errors were present on screen)
- Included `Payment details` in a `ScrollView` so that everything is visible when font size is large (previously fields were overlapping when keyboard and errors were shown)

IPC-178


health-sdk/sdk/src/main/res/values/strings.xml
```diff
@@ -52,4 +52,8 @@
     <string name="ghs_more_information_fragment_title">Mehr Informationen</string>
     <string name="ghs_privacy_policy">Datenschutzerklärung</string>
     <string name="ghs_privacy_policy_link_url"/>
+
+    //TODO get German wording and include overridable strings in documentation
```
danicretu, line 56: @a-szotyori , here are the strings for which German wording is needed. I added the `Invalid Amount` string just so that the `TalkBack` feature would speak out something more meaningful than just `Invalid`.

a-szotyori: Thank you, @danicretu for highlighting what needs to be translated. Here are the German translations:
* ghs_close_content_description -> "Schließen"
* ghs_invalid_amount_content_description -> "Ungültiger Bezahlbetrag"

a-szotyori (approved):
Thank you!

PR #414, opened 2024-04-04
Ipc 176 accessibility payment component

danicretu:
- Switched to using `lineSpacingMultiplier` rather than `lineHeight`
- Added content description for `MoreInformation` button


a-szotyori (approved):
Thank you!

PR #415, opened 2024-04-04
Ipc 179 accessibility payment info

danicretu:
- Added `contentDescription` for Gini logo
- Added `contentDescription` for payment provider logos


health-sdk/sdk/src/main/res/values/strings.xml
```diff
@@ -52,4 +52,6 @@
     <string name="ghs_more_information_fragment_title">Mehr Informationen</string>
     <string name="ghs_privacy_policy">Datenschutzerklärung</string>
     <string name="ghs_privacy_policy_link_url"/>
+
+    <string name="ghs_payment_provider_logo_content_description">logo</string>
```
danicretu, line 56: @a-szotyori , I added this so we can have a content description for the payment provider icons list. Since we don't currently have a large icon for the payment provider, it doesn't make sense to me to make the icons bigger when the font is increased. Perhaps when we will have larger icons we can increase the size of the logos as well.

Thoughts?

a-szotyori: I agree. The important thing is that text size increases. The size of the logos themselves we don't need to change ourselves when the font size changes. I quickly checked that Android will increase their size automatically when users increase the "display size" so we are fine leaving the resizing to Android:
Screenshot_20240404_203812

danicretu: Thanks for confirming! Then all that's left for this to be ready for review is the translation for `logo`.  I couldn't think of a more descriptive word, so any other suggestions are welcome.

a-szotyori (approved):
Thank you!

PR #424, opened 2024-04-19
Ipc 240 install app bottom sheet

danicretu:
Added `InstallAppBottomSheetFragment`, which will prompt the user to install the selected bank app. 
Refactored `ReviewFragment` a bit so we can send `paymentComponent` forward to the new fragment.


a-szotyori (commented):
Thank you, Daniela! The PR approved from my point of view. 👍 Let's wait until Miguel and István confirm the texts in Figma and then update the strings before moving on to QA.

a-szotyori (approved):
Thank you! 👏

PR #425, opened 2024-04-22
Ipc 240 data flow refactor

a-szotyori:
Refactors the data flow of the selected payment provider app by modifying:
- the PaymentComponent to update the selected app when rechecking the installed state
- the review fragment and the "install app bottom sheet" to use PaymentComponent's `selectedPaymentProviderAppFlow` as the source of truth for the payment provider app

This way the data flow is reactive: "recheck installed state" event is sent to PaymentComponent which then emits an updated selected payment provider app downstream to the review fragment and the "install app bottom sheet". These then update their UI and state accordingly.


danicretu (approved):
It's more elegant and simplified indeed. Thanks for the refactoring 👍

PR #430, opened 2024-05-07
Pp 438 user journey main structure

abolfazlimahdi:
This is the main structure of integrating an analytics tool.


a-szotyori (commented):
Looks good, Mahdi! I only left suggestions and a clarification.

a-szotyori:
@mahdiabolfazli Regarding the failing tests, I see they are emulator related. You can apply these [changes](https://github.com/gini/gini-mobile-android/commit/acfb500614ad2bdcb48b04544e42c9cdcb1bc971) to the failing workflows to switch to using Ubuntu runners with KVM.

a-szotyori (commented):
Looks good! I just have one more thing to discuss 🙈.

What do you think of merging all analytics related changes into a dedicated branch called something like `PP-354-analytics-feature` so that `main` is not affected until the analytics feature is fully complete, QA-d and ready for release? Just to avoid putting incomplete work into `main`. This would allow us to do any maintenance or bugfix work on `main` as usual while in parallel you develop the analytics feature.

abolfazlimahdi:
> you
Thanks for reviewing and for the comment. I will not merge it into the main branch until it is fully developed and only branch from the previous branch and so on... I picked up this to be aligned with iOS. However, I see that we can stick to our previous process. I will continue working on one branch until it is completed. Just ask you for you review after each ticket is completed. Is it ok?

a-szotyori:
@mahdiabolfazli Yes, thank you, Mahdi! Then just re-request reviews when you complete a ticket. 👍

a-szotyori (approved):

PR #433, opened 2024-05-13
Ipc 251 android health api library download pdf

danicretu:
- Creates temp PDF file
- Shares it via intent chooser

This PR will need to be altered once BE is ready and we receive the payment request document from the API. I will fix the failing tests related to BankSDK and HealthApiLibrary then, as some of them are related to the placeholders I added to be able to continue with development.


health-sdk/sdk/src/main/java/net/gini/android/health/sdk/review/ReviewFragment.kt
```diff
@@ -547,25 +554,34 @@ class ReviewFragment private constructor(
...
-        startActivity(Intent.createChooser(shareIntent, ""))
+        startActivity(Intent.createChooser(shareIntent, uriForFile.lastPathSegment))
     }
 
     private fun handlePaymentNextStep(paymentNextStep: ReviewViewModel.PaymentNextStep) {
         when (paymentNextStep) {
-            ReviewViewModel.PaymentNextStep.OpenSharePdf -> startSharePdfIntent()
+            ReviewViewModel.PaymentNextStep.DownloadPaymentRequestFile -> {
+                binding.loading.isVisible = true
+                viewModel.getFileAsByteArray(requireContext().externalCacheDir)
```
a-szotyori, line 583: [Suggestion] I think we are mixing the responsibilities of the fragment and the viewmodel. Currently the code flow for getting the pdf is (ignoring the "open with" onboarding sheet for simplicity):
1. Fragment: notify ViewModel that payment button was tapped
2. ViewModel: notify Fragment to download the pdf
3. Fragment: notify ViewModel to download the pdf
4. ViewModel: notify Fragment to share the pdf

Could the flow be simplified to:
1. Fragment: notify ViewModel that payment button was tapped
2. ViewModel: 
  a. notify Fragment to show loading indicator for pdf download
  b. download pdf
  c. notify Fragment to hide loading indicator for pdf download
  d. notify Fragment to share the pdf OR to show error

health-sdk/sdk/src/main/java/net/gini/android/health/sdk/review/ReviewViewModel.kt
```diff
@@ -50,6 +60,11 @@ internal class ReviewViewModel(val giniHealth: GiniHealth, val configuration: Re
     private val _paymentProviderApp = MutableStateFlow<PaymentProviderApp?>(null)
     val paymentProviderApp: StateFlow<PaymentProviderApp?> = _paymentProviderApp
 
+    private val _paymentNextStep = MutableSharedFlow<PaymentNextStep>(
+        extraBufferCapacity = 1,
+        onBufferOverflow = BufferOverflow.DROP_OLDEST)
```
a-szotyori, line 65: [Clarification] Can you help me please to understand why buffering is needed?

a-szotyori (changes requested):
Great work! I have one change request and some discussion points 🤓

health-sdk/sdk/src/main/java/net/gini/android/health/sdk/util/extensions/File.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import java.io.File
+import java.io.FileOutputStream
+import java.io.IOException
+
+fun File.createTempPdfFile(byteArray: ByteArray, fileName: String): File? {
+    val file = File("${this.path}/", "${fileName}.pdf")
+
+    try {
+        val os = FileOutputStream(file, false)
```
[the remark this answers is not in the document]
danicretu: Thank for pointing this out - changed the block of code to `use`.

health-sdk/sdk/src/main/java/net/gini/android/health/sdk/review/ReviewFragment.kt
danicretu: Yes, I was overcomplicating things a bit...simplified the flow, but sending `externalCacheDir` as a param when tapping on payment button.

health-sdk/sdk/src/main/java/net/gini/android/health/sdk/review/ReviewViewModel.kt
danicretu: The `extraBufferCapacity` is just so the flow emits events. If we remove it, then nothing is emitted. 
I've removed `onBufferOverflow` as it was not adding any value - originally I thought it was helping with events to not be collected again on configuration change, but it seems to be working without it.

a-szotyori (changes requested):
Thank you for simplifying the flow between fragment and viewmodel! I have only some small remaining requests. 🙈

health-sdk/sdk/src/main/java/net/gini/android/health/sdk/review/ReviewViewModel.kt
```diff
@@ -271,20 +272,26 @@ internal class ReviewViewModel(val giniHealth: GiniHealth, val configuration: Re
...
             if (paymentProviderApp.value?.isInstalled() == true) _paymentNextStep.tryEmit(PaymentNextStep.RedirectToBank)
             else _paymentNextStep.tryEmit(PaymentNextStep.ShowInstallApp)
             return
         }
-        if (openWithCounter < SHOW_OPEN_WITH_TIMES) _paymentNextStep.tryEmit(PaymentNextStep.ShowOpenWithSheet) else _paymentNextStep.tryEmit(PaymentNextStep.DownloadPaymentRequestFile)
+        if (openWithCounter < SHOW_OPEN_WITH_TIMES) _paymentNextStep.tryEmit(PaymentNextStep.ShowOpenWithSheet)
+        else {
+            _paymentNextStep.tryEmit(PaymentNextStep.SetLoadingVisibility(true))
+            getFileAsByteArray(externalCacheDir)
+        }
```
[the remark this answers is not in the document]
danicretu: Yes, my bad - an oversight on my part.

health-sdk/sdk/src/main/java/net/gini/android/health/sdk/review/ReviewFragment.kt
```diff
@@ -574,6 +589,10 @@ class ReviewFragment private constructor(
             }
             ReviewViewModel.PaymentNextStep.ShowOpenWithSheet -> viewModel.paymentProviderApp.value?.let { showOpenWithDialog(it) }
             ReviewViewModel.PaymentNextStep.ShowInstallApp -> viewModel.paymentProviderApp.value?.let { showInstallAppDialog(it) }
+            is ReviewViewModel.PaymentNextStep.OpenSharePdf -> {
+                binding.loading.isVisible = false
```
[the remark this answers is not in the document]
danicretu: It's still needed here, as the visibility isn't set to GONE from the viewModel. I opted for this in case there is any extra functionality that we might want to add in the UI, so the UI layer decides when to hide the loading indicator.

I added the `setLoadingVisibility` step in the VM because it's the VM which decides what the next step should be, based on all the variables that we monitor. It currently sends a `boolean` parameter to the `sharedFlow` but it's always `true` at the moment. I opted for this for the same reason as I mentioned above, if there's any case in which we should hide the loading from the VM at any point in the future (so we don't need to refactor). If you consider this to be an overworking of a simple thing, please let me know and I can switch it over to a plain `object`.

health-sdk/sdk/src/main/java/net/gini/android/health/sdk/review/ReviewViewModel.kt
```diff
@@ -295,12 +294,26 @@ internal class ReviewViewModel(val giniHealth: GiniHealth, val configuration: Re
...
+                    GiniHealth.PaymentState.Error(throwable)
+                    _paymentNextStep.tryEmit(PaymentNextStep.OpenSharePdfError(throwable.message ?: "Error getting payment request"))
+                    return@withContext
+                }
                 val byteArrayResource = async {  giniHealth.giniHealthAPI.documentManager.getPaymentRequestDocument(paymentRequest.id) }.await()
                 when (byteArrayResource) {
-                    is Resource.Cancelled -> throw Exception("Cancelled")
-                    is Resource.Error -> throw Exception(byteArrayResource.exception)
+                    is Resource.Cancelled -> {
+                        GiniHealth.PaymentState.Error(Exception("Cancelled"))
```
a-szotyori, line 308: [Request] This event needs to be emitted on `giniHealth.setOpenBankState()` so that clients can receive it. It is also collected by the ReviewFragment and I have a proposal how to differentiate in the ReviewFragment between "payment request error" and "pdf error" and avoid showing both:
- Emit all pdf errors as a `GiniHealth.PaymentState.Error`, but use a custom Exception subclass, for example `OpenSharedPdfException`. This way we have one error channel.
- Check the instance of the Throwable in ReviewFragment's `handlePaymentState()` method and show the pdf error message if it's the `OpenSharedPdfException`.

health-sdk/sdk/src/main/java/net/gini/android/health/sdk/review/ReviewViewModel.kt
```diff
@@ -295,12 +294,26 @@ internal class ReviewViewModel(val giniHealth: GiniHealth, val configuration: Re
...
+                        _paymentNextStep.tryEmit(PaymentNextStep.OpenSharePdfError("Cancelled"))
+                    }
+                    is Resource.Error -> {
+                        GiniHealth.PaymentState.Error(Exception(byteArrayResource.exception))
+                        _paymentNextStep.tryEmit(PaymentNextStep.OpenSharePdfError(byteArrayResource.exception?.message ?: "Error getting payment request file"))
```
a-szotyori, line 313: [Request] Same as in the comment for the Cancelled case.

a-szotyori (changes requested):
I'm sorry, Daniela, for requesting changes again. It's crucial that we get this right.

health-sdk/sdk/src/main/java/net/gini/android/health/sdk/review/ReviewViewModel.kt
danicretu: As discussed, for now both `payment request` and `get pdf` will display the same error and follow the same steps after tapping on `Retry`. Should we need separate error messages, an `OpenSharedPdfException` will be introduced to allow us to differentiate between the two.

a-szotyori (approved):
Thank you, Daniela! Looks great now! Regarding the error messages there [might be](https://ginis.atlassian.net/browse/IPC-229?focusedCommentId=24449) one more change needed, but we have to wait for Miguel.

a-szotyori (approved):
Thank you!

PR #435, opened 2024-05-14
Ipc 260 open with and gpc supported

danicretu:
- Added support for gpcSupportedPlatforms and `openWithSupportedPlatforms`
- Add filtering in `HealthApiDocumentRepository`

### DIFF
--- a/health-api-library/library/src/main/java/net/gini/android/health/api/models/PaymentProvider.kt
+++ b/health-api-library/library/src/main/java/net/gini/android/health/api/models/PaymentProvider.kt
@@ -66,7 +66,7 @@ data class PaymentProvider(
         return result
     }
 
-    fun gpcSupported(): Boolean = gpcSupportedPlatforms.map { it.lowercase(Locale.getDefault()) }.contains(ANDROID_PLATFORM)
+    fun gpcSupported(): Boolean = gpcSupportedPlatforms.map { it.lowercase(Locale.ENGLISH) }.contains(ANDROID_PLATFORM)
 
     /**
      * A payment provider's color scheme.
--- a/health-api-library/library/src/main/java/net/gini/android/health/api/models/PaymentProvider.kt
+++ b/health-api-library/library/src/main/java/net/gini/android/health/api/models/PaymentProvider.kt
@@ -1,6 +1,7 @@
 package net.gini.android.health.api.models
 
 import net.gini.android.health.api.response.PaymentProviderResponse
+import java.util.Locale
 
 /**
  * A payment provider is a Gini partner which integrated Gini Pay Connect (via Gini Bank SDK) into their mobile apps.
@@ -64,7 +65,7 @@ data class PaymentProvider(
         return result
     }
 
-    fun gpcSupported(): Boolean = gpcSupportedPlatforms.contains("android")
+    fun gpcSupported(): Boolean = gpcSupportedPlatforms.map { it.lowercase(Locale.getDefault()) }.contains(ANDROID_PLATFORM)
 
     /**
      * A payment provider's color scheme.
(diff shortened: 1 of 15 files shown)


health-api-library/library/src/main/java/net/gini/android/health/api/response/PaymentProviderResponse.kt
```diff
@@ -12,8 +14,13 @@ internal data class PaymentProviderResponse(
     @Json(name = "colors") val colors: Colors,
     @Json(name = "iconLocation") val iconLocation: String,
     @Json(name = "playStoreUrlAndroid") val playStoreUrl: String?,
-    @Json(name = "gpc_supported") val gpcSupported: Boolean?,
-)
+    @Json(name = "gpcSupportedPlatforms") val gpcSupportedPlatforms: List<String>?,
+    @Json(name = "openWithSupportedPlatforms") val openWithSupportedPlatforms: List<String>?
+) {
+
+    fun isEnabled(): Boolean = (gpcSupportedPlatforms?.map { it.lowercase(Locale.getDefault()) }?.contains(ANDROID_PLATFORM) == true )
```
a-szotyori, line 21: [Request] `Locale.getDefault()` is dangerous here because it can happen that the default locale does not convert to lowercase in the way we expect it. I [found that using](https://stackoverflow.com/a/10336915) `Locale.ENGLISH` is safe for identifiers.

health-sdk/sdk/src/main/java/net/gini/android/health/sdk/paymentcomponent/PaymentComponent.kt
```diff
@@ -88,7 +88,8 @@ class PaymentComponent(private val context: Context, private val giniHealth: Gin
                     LOG.debug("Loaded payment providers")
                     LOG.debug("Loading installed payment provider apps")
 
-                    val paymentProviderApps = getPaymentProviderAppsSorted(paymentProvidersResource.data)
+                    val supportedPaymentProviderApps = paymentProvidersResource.data.filter { it.gpcSupportedPlatforms?.contains("android") == true || it.openWithSupportedPlatforms?.contains("android") == true }
```
a-szotyori, line 91: [Request] This is not needed now that you are filtering using `PaymentProvider.isEnabled()`.

health-api-library/library/src/main/java/net/gini/android/health/api/models/PaymentProvider.kt
```diff
@@ -61,6 +66,8 @@ data class PaymentProvider(
         return result
     }
 
+    fun gpcSupported(): Boolean = gpcSupportedPlatforms.map { it.lowercase(Locale.getDefault()) }.contains(ANDROID_PLATFORM)
```
a-szotyori, line 69: [Request] We should use `Locale.ENGLISH` here as well.

a-szotyori (changes requested):
Thank you, Daniela, nicely done! I only have three small requests.

health-api-library/library/src/main/java/net/gini/android/health/api/response/PaymentProviderResponse.kt
danicretu: Thank you for pointing this out - changed it to use `Locale.ENGLISH`

health-sdk/sdk/src/main/java/net/gini/android/health/sdk/paymentcomponent/PaymentComponent.kt
danicretu: Yes, forgot about leaving this in - removed it.

a-szotyori (approved):

PR #437, opened 2024-05-20
IPC-261 Previously extracted information visible on `InvoicesActivity`

danicretu:
IPC-261

### DIFF
--- a/health-sdk/example-app/src/main/java/net/gini/android/health/sdk/exampleapp/invoices/ui/InvoicesActivity.kt
+++ b/health-sdk/example-app/src/main/java/net/gini/android/health/sdk/exampleapp/invoices/ui/InvoicesActivity.kt
@@ -8,7 +8,6 @@ import android.view.View
 import android.view.ViewGroup
 import android.widget.LinearLayout
 import android.widget.TextView
-import androidx.activity.OnBackPressedCallback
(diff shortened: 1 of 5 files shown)


health-sdk/example-app/src/main/java/net/gini/android/health/sdk/exampleapp/invoices/ui/InvoicesActivity.kt
```diff
@@ -130,6 +130,16 @@ class InvoicesActivity : AppCompatActivity() {
...
+                    viewModel.openBankState.collect { paymentState ->
+                        when (paymentState) {
+                            is GiniHealth.PaymentState.Success -> {
+                                viewModel.updateDocument()
+                            }
+                            else -> {}
+                        }
+                    }
+                }
             }
```
danicretu, line 143: The reason for keeping this in is the following use case:
1. The users reaches `ReviewFragment`
2. The user modifies the amount and / or the recipient
3. The user taps on `To the banking app`
4. From the bank screen the user hits `Back` until reaching the `Invoices screen` (without completing the payment)

The `InvoiceActivity` is still active at this point, so we can just update the one document instead of adding a whole reload to the resumed state of the lifecycle. I am open to changes though 😁

a-szotyori (approved):
Thank you!

PR #440, opened 2024-05-28
Pp 470 change implementation of the back navigation on android

danicretu:
Testing if making an edit here triggers our workflows.


capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragmentImpl.java
```diff
@@ -371,10 +372,7 @@ private void handleOnBackPressed() {
         activity.getOnBackPressedDispatcher().addCallback(mFragment.getViewLifecycleOwner(), new OnBackPressedCallback(true) {
             @Override
             public void handleOnBackPressed() {
-                trackCameraScreenEvent(CameraScreenEvent.EXIT);
-                setEnabled(false);
-                remove();
-                mFragment.getActivity().getOnBackPressedDispatcher().onBackPressed();
+                onBackPressed();
```
a-szotyori, line 375: [Request] Here we should only track the exit event because the navigation is handled by the fragment manager / navigation component:
```suggestion
                trackCameraScreenEvent(CameraScreenEvent.EXIT);
                setEnabled(false);
                remove();
```

It might have unexpected side-effects if we pop the back stack and then under the hood the fragment manager or the navigation component pops it again.

capture-sdk/sdk/src/main/java/net/gini/android/capture/analysis/AnalysisFragmentImpl.java
```diff
@@ -283,14 +282,21 @@ private void handleOnBackPressed() {
         activity.getOnBackPressedDispatcher().addCallback(mFragment.getViewLifecycleOwner(), new OnBackPressedCallback(true) {
             @Override
             public void handleOnBackPressed() {
-                trackAnalysisScreenEvent(AnalysisScreenEvent.CANCEL);
                 setEnabled(false);
                 remove();
-                mFragment.getActivity().getOnBackPressedDispatcher().onBackPressed();
+                onBack();
```
a-szotyori, line 287: [Request] Here as well we should only track the cancel event:
```suggestion
                trackAnalysisScreenEvent(AnalysisScreenEvent.CANCEL);
                setEnabled(false);
                remove();
```

capture-sdk/sdk/src/main/java/net/gini/android/capture/review/multipage/MultiPageReviewFragment.java
```diff
@@ -137,12 +139,9 @@ public void onCreate(final Bundle savedInstanceState) {
         OnBackPressedCallback callback = new OnBackPressedCallback(true /* enabled by default */) {
             @Override
             public void handleOnBackPressed() {
-                trackReviewScreenEvent(ReviewScreenEvent.BACK);
+                onBack();
```
a-szotyori, line 142: [Request] Here as well we should only track the back event:
```suggestion
                trackReviewScreenEvent(ReviewScreenEvent.BACK);
                setEnabled(false);
                remove();
```

capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragment.java
```diff
@@ -65,10 +68,12 @@ private void setListener(@NonNull final CameraFragmentImpl fragmentImpl, @Nullab
                             + "You can set it with CameraFragmentCompat#setListener() or "
                             + "by making the host activity implement the CameraFragmentListener.");
         }
+
+        mCancelListener = cancelListener;
```
a-szotyori, line 72: [Request] I think this is not needed because it's in a private method which always receives the `mCancelListener` so this just boils down to `mCancelListener = mCancelListener`. Also then the method can be renamed back to singular `setListener()`.

capture-sdk/sdk/src/main/java/net/gini/android/capture/error/ErrorFragment.kt
```diff
@@ -60,11 +62,16 @@ class ErrorFragment : Fragment(), FragmentImplCallback {
         return fragmentImpl.onCreateView(inflater, container, savedInstanceState)
     }
 
-    fun setListener(
-        listener: EnterManuallyButtonListener
+    fun setListeners(
+        listener: EnterManuallyButtonListener,
```
a-szotyori, line 66: [Request] This can remain singular: `setListener`

a-szotyori (changes requested):
Awesome work! I have only some small requests.

capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragmentImpl.java
danicretu: In this case, the `CameraFragmentImpl` overrides `handleOnBackPressed` behavior by adding its own callback. `remove` only removes the callback, but does not execute the back press event. Because of this,
not calling our own `onBackPressed` method does not exit the camera flow. The user would need to tap a second time on the back button in order to exit it (because then the custom back callback would have been removed).

capture-sdk/sdk/src/main/java/net/gini/android/capture/analysis/AnalysisFragmentImpl.java
danicretu: The same comment as for `CameraFragmentImpl`.

capture-sdk/sdk/src/main/java/net/gini/android/capture/review/multipage/MultiPageReviewFragment.java
danicretu: Here as well, same comment as above.

capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragment.java
danicretu: Good catch, thank you for pointing it out!

capture-sdk/sdk/src/main/java/net/gini/android/capture/analysis/AnalysisFragmentImpl.java
a-szotyori: Oh, I see! Then no changes needed. 👍

capture-sdk/sdk/src/main/java/net/gini/android/capture/review/multipage/MultiPageReviewFragment.java
a-szotyori: Oh, I see! Then no changes needed. 👍

a-szotyori (approved):
Thank you! To save on overhead I will prepare the release candidate in this branch (bump the version number) and then upload the RC build for testing. So from now on we can treat this branch as a release candidate branch:
- in case Syeda finds issues they can be fixed here
- after all is green I will add release tags to trigger the release workflows
- after those pass I will merge this into main

danicretu:
Understood, thank you!

PR #443, opened 2024-05-28
feat(bank-sdk): Add return assistant user analytics events 

ndubkov-distcotech:
Ticket: PP-446


abolfazlimahdi (approved):
Thank you. Everything looks good now.

PR #446, opened 2024-06-05
feat(capture-sdk): Add camera access permission screen analytics

ndubkov-distcotech:
Ticket: PP-447


capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragmentImpl.java
```diff
@@ -1902,4 +1912,28 @@ private void onBackPressed() {
             mCancelListener.onCancelFlow();
         }
     }
+
+    private void trackCameraNoPermissionMessageShownEvent() {
```
abolfazlimahdi, line 1916: [Suggestion] I think we don't need to add functions for one line actions. What do you think? If we need them, I would suggest better names. This is the CameraAccess permission screen, not the NoPermissionMessage ....

abolfazlimahdi (commented):
I have added a comment in the PR. I would also suggest that you send the status of the permission. We can talk about it if you want. Thanks :-)

capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragmentImpl.java
ndubkov-distcotech: Resolved

abolfazlimahdi (approved):
Thank you!

PR #447, opened 2024-06-06
Ec 62 repackage health sdk into merchant sdk

danicretu:
TODO: update Figma customization guide links when we have them

- editing to trigger checks


a-szotyori (approved):
Thank you, Daniela! I just updated a few Merchant SDK descriptions and removed the outdated migration guide

merchant-sdk/example-app/README.md
```diff
@@ -1,11 +1,11 @@
 Gini Merchant SDK Example App
-=========================
+=============================
 
 This example app provides you with a sample usage of the Gini Merchant SDK.
 
-The Gini Health API Library is used for analyzing documents and creating payment requests.
+The Gini Merchant API Library is used for analyzing documents and creating payment requests.
```
danicretu, line 6: Hi @a-szotyori , should we change `Health API` to `Merchant API` at this point? Or only after we duplicate? 

Not sure clients need to know how it's working under the hood, so your approach could be better. Just asking because I've intentionally left it as `Health API` everywhere, and would need to go back and change it.

a-szotyori: Thank you for pointing this out! So as far as I know go-live will be with using `https://merchant-api.gini.net`, but through the Health API Library by changing the base url to `merchant-api.gini.net`. 

Without refactoring we can't hide the Health API Library from client devs because it needs to be instantiated for [authentication](https://developer.gini.net/gini-mobile-android/health-sdk/sdk/4.1.0/html/authentication.html) and passed to `GiniMerchant`.

I will change the name of Merchant API Library back to Health API Library. I did the renaming too quickly without thinking about its effects.

merchant-sdk/example-app/src/main/java/net/gini/android/merchant/sdk/exampleapp/upload/UploadViewModel.kt
```diff
@@ -73,11 +72,12 @@ class UploadViewModel(
         }
     }
 
-    private fun setDocumentForReview(documentId: String) {
-        viewModelScope.launch {
-            giniMerchant.setDocumentForReview(documentId)
-        }
-    }
+    // TODO EC-62: we don't expose the document anymore in the Merchant SDK and instead we pass the document id to the ReviewActivity
```
danicretu, line 75: This has been changed in EC-60, we set the document for review in the MerchantSDK, when instantiating the container fragment. I don't think the client needs to handle this step, do you think otherwise? 🤔

merchant-sdk/example-app/src/main/java/net/gini/android/merchant/sdk/exampleapp/review/ReviewActivity.kt
```diff
@@ -163,6 +162,8 @@ class ReviewActivity : AppCompatActivity() {
                 return@launch
             }
 
+            viewModel.giniMerchant.setDocumentForReview(documentId)
```
danicretu, line 165: I think this can be removed when merging with EC-60 - since the instantiation of the container fragment contains the documentId, setting the document for review can be done internally.

merchant-sdk/example-app/src/main/java/net/gini/android/merchant/sdk/exampleapp/review/ReviewActivity.kt
danicretu, line 1: Will this be needed in the MerchantSDK? Will clients be able to import / take a photo?

merchant-sdk/example-app/src/main/java/net/gini/android/merchant/sdk/exampleapp/MainViewModel.kt
```diff
@@ -38,9 +38,10 @@ class MainViewModel(
         ++currentIndex
     }
 
-    fun setDocumentForReview(documentId: String) {
```
danicretu: I think this can be removed when merging with EC-60 - since the instantiation of the container fragment contains the documentId, setting the document for review can be done internally.

merchant-sdk/example-app/src/main/java/net/gini/android/merchant/sdk/exampleapp/review/ReviewActivity.kt
a-szotyori: It will be needed for testing. We have to refactor how it works under the hood to pass the file or image to the Merchant SDK where we do the upload ourselves to get the payment details. This way we can also stop using Health API Library directly here.

**Sidenote:**
This makes me realise that we also have to refactor the Invoices List to have an orders list instead 🤔. So the idea I came up with quickly is (and this just the starting point, we need to include Miguel and Nadya in the discussion):
- we replace the hardcoded test health invoices with e-commerce invoices
- instead of the "upload invoices" button to add items to the list we hardcode the list with the test e-commerce invoices
- we add a small "pay" button on each list item which then launches the Merchant SDK with the e-commerce invoice
- we keep the list static meaning that the shown data is never updated and the list is not extended when importing invoices or taking a photo

a-szotyori: Addition to the sidenote: OR we remove the Invoices List and just use the import option to show the fake merchant app screen (with the payment details) from where the Merchant SDK can be launched.

merchant-sdk/example-app/src/main/java/net/gini/android/merchant/sdk/exampleapp/upload/UploadViewModel.kt
a-szotyori: Thank you, I will delete it then completely.

merchant-sdk/example-app/src/main/java/net/gini/android/merchant/sdk/exampleapp/review/ReviewActivity.kt
a-szotyori: I agree 👍

merchant-sdk/example-app/src/main/java/net/gini/android/merchant/sdk/exampleapp/MainViewModel.kt
a-szotyori: Thank you, I will delete it then completely.

merchant-sdk/example-app/src/main/java/net/gini/android/merchant/sdk/exampleapp/review/ReviewActivity.kt
danicretu: I agree with the suggestion, I think the option with a `pay button` will be more self explanatory for the example app.

Apart from transferring the upload responsibilities to the sdk instead of the example app, when going via the `Take Photo` or `Upload` route, upon tapping on the `Pay Button` should we start the MerchantSDK directly and show the bottom sheet with the `paymentComponent`? 

Or also show the `invoiceDetail` screen that we currently show when selecting an invoice form the `Invoices List`?

a-szotyori: I suggested in our [slack channel](https://ginis.slack.com/archives/C0736AKHH0T/p1718790245972319) to show the `invoiceDetail` screen with fake details. To simply reuse the `invoiceDetail` screen and avoid launching the Merchant SDK from different places.

PR #449, opened 2024-06-07
Pp 354 user analytics feature

abolfazlimahdi:


capture-sdk/sdk/src/main/java/net/gini/android/capture/tracking/useranalytics/UserAnalyticsScreen.kt
```diff
@@ -0,0 +1,41 @@
...
+    ERROR("error"),
+    ONBOARDING_FLAT_PAPER("onboarding_flat_paper"),
+    ONBOARDING_LIGHTING("onboarding_lighting"),
+    ONBOARDING_MULTIPLE_PAGES("onboarding_multiple_pages"),
+    ONBOARDING_QR_CODE("onboarding_qr_code"),
+    ONBOARDING_CUSTOM_1("onboarding_custom_1"),
+    ONBOARDING_CUSTOM_2("onboarding_custom_2"),
+    ONBOARDING_CUSTOM_3("onboarding_custom_3"),
+    ONBOARDING_CUSTOM_4("onboarding_custom_4"),
+    ONBOARDING_CUSTOM_5("onboarding_custom_5"),
```
ndubkov-distcotech, line 21: [Request]
We need to refactor this part because there is can be more than 5 custom pages in onboarding 
(Clarified this with Valentina)

capture-sdk/sdk/src/main/java/net/gini/android/capture/tracking/useranalytics/UserAnalyticsScreen.kt
```diff
@@ -0,0 +1,41 @@
...
+    RETURN_ASSISTANT("return_assistant"),
+    EDIT_RETURN_ASSISTANT("edit_return_assistant")
+}
+
+
+fun getOnboardingScreenNameForUserAnalytics(id: Int): UserAnalyticsScreen {
+    return when (id) {
+        R.string.gc_onboarding_qr_code_title -> UserAnalyticsScreen.ONBOARDING_QR_CODE
```
ndubkov-distcotech, line 29: [Request] I think it will be better to do this check inside of fragment. 
The analytics is a separate "module" of the project so it will be better to keep it isolated from View layer. 
In my opinion, this functiuon should be splitted into 2 separate functions: first one to map based on `StringRes` and 2nd one can be removed at all after refactoring from comment above

ndubkov-distcotech (changes requested):

ndubkov-distcotech (approved):

PR #453, opened 2024-06-11
feat(bank-sdk): Add SDK open/closed events

ndubkov-distcotech:
Ticket: PP-517


abolfazlimahdi (changes requested):
The sdk_closed event is only sent with RA files. Could not see any events for "upload a valid invoice document", taking a picture from a valid invoice", and "scanning a valid qr-code". Thank you :-)

ndubkov-distcotech:
> The sdk_closed event is only sent with RA files. Could not see any events for "upload a valid invoice document", taking a picture from a valid invoice", and "scanning a valid qr-code". Thank you :-)

Just checked and its sent as it should event if return assistant is disabled

abolfazlimahdi (approved):
Everything looks great now. Thank you!

PR #454, opened 2024-06-11
Ipc 273 add payment endpoint and document expiration date

a-szotyori:
- Move the get payment resolution endpoint from Bank API Library into the Core API Library so that it's also available in the Health API Library.
- Parse the optional document expiration date which is provided by the Health API Library.


danicretu (approved):

danicretu (approved):
Thank you!  👌

PR #460, opened 2024-06-17
Ec 60 configurability of review screen

danicretu:
Please refer to commit details for more details.

Open to changes to the new class names, perhaps I didn't choose the most self explanatory names 🤔


merchant-sdk/sdk/src/main/java/net/gini/android/merchant/sdk/integratedFlow/MerchantFragment.kt
```diff
@@ -0,0 +1,270 @@
...
+    val isAmountFieldEditable: Boolean = false
+): Parcelable
+
+/**
+ * The [MerchantFragment] provides a container for all screens that should be displayed for the user
+ * during the payment process (eg. [PaymentComponentBottomSheet], [BankSelectionBottomSheet], [ReviewFragment]).
+ *
+ * It handles the display logic for all screens. A new instance can be created using the [PaymentComponent.getContainerFragment] method.
+ */
+class MerchantFragment private constructor(
```
a-szotyori, line 73: @danicretu I am sorry for making you change the name again, but [naming is hard](https://martinfowler.com/bliki/TwoHardThings.html)... I think we should not name it MerchantFragment. Reading this makes me think of some UI related to a merchant (like viewing some details about the merchant). Knowing that it actually is about performing a payment makes me think that this is name is not a good fit at all 😞 

What do you think about `PaymentFlowFragment`? It reflects that it is about a payment and that it manages the whole flow.

merchant-sdk/sdk/src/main/java/net/gini/android/merchant/sdk/integratedFlow/MerchantFragment.kt
```diff
@@ -0,0 +1,270 @@
...
+     */
+    val shouldHandleErrorsInternally: Boolean = true,
+
+    /**
+     * If set to `true`, the [Amount] field of shown as part of the [ReviewFragment] will be editable.
+     * If set to `false`, the [Amount] field will be read-only.
+     *
+     * Default value is `false`
+     */
+    val isAmountFieldEditable: Boolean = false
```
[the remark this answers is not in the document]
danicretu: Thanks for pointing this out - will change it. 👍

merchant-sdk/example-app/src/main/java/net/gini/android/merchant/sdk/exampleapp/invoices/ui/InvoicesViewModel.kt
```diff
@@ -56,47 +52,24 @@ class InvoicesViewModel(
...
-
-            _paymentReviewFragmentFlow.value = PaymentReviewFragmentState.Loading
+    fun setSelectedInvoiceItem(invoiceItem: InvoiceItem) = viewModelScope.launch {
+        _selectedInvoiceItem.emit(invoiceItem)
+    }
 
-            val documentWithExtractions =
-                invoicesRepository.invoicesFlow.value.find { it.documentId == documentId }
+    fun startIntegratedPaymentFlow(documentId: String) {
+        _startIntegratedPaymentFlow.tryEmit(giniMerchant.getContainerFragment(documentId, paymentFlowConfiguration))
```
a-szotyori, line 64: [Suggestion] Since `getContainerFragment()` is synchronous we could avoid using the `startIntegratedPaymentFlow` and just return the fragment directly here.

danicretu: `InvoicesViewModel` is currently shared between `InvoicesActivity` and `InvoiceDetailsFragment`. The reason why I opted for a flow here is because the payment flow is started by tapping on the `Pay Now` button in `InvoiceDetailsFragment`, but `InvoicesActivity` is the one that handles adding and removing fragments from its container view. 
So `InvoiceDetailsFragment` -> asks for the fragment -> sends it over to `InvoicesActivity` -> shows it in its hierarchy.
Some sort of flow is needed in my opinion to signal to the activity that is needs to do its job, but it can be a simple void flow...and let the activity call `giniMerchant.getContainerFragment`. Would this be a simpler / more straightforward approach in your opinion?

a-szotyori: Thank you for the clarification. In that case it can remain as it is. 

Sidenote: in EC-15 we will have to [replace Invoices List and Invoice Details](https://ginis.slack.com/archives/C0736AKHH0T/p1719237916687839?thread_ts=1718790245.972319&cid=C0736AKHH0T) with an Orders Screen and an Order Details Screen. We could then have an OrdersActivity which uses Jetpack Navigation to manage the navigation between Orders Screen <-> Order Details Screen <-> PayFlowFragment. But we don't have to decide that now.

a-szotyori (approved):
Thank you!

PR #461, opened 2024-06-17
feat(health-sdk): Fixed error not showing in 'share with' scenario

danicretu:
IPC-279

I opted not to change `StateFlow` to `SharedFlow` as I found that adding an imperceptible delay when emitting 2 consecutive events fixed the issue. It's a smaller change in my opinion, with less impact on the codebase. I'm open to discussions if switching to `SharedFlow` is the preferred route.

Please note: when tapping on `Retry`, the snackbar flickers when being dismissed. It **might** be related to [this](https://github.com/material-components/material-components-android/issues/648) bug that has apparently been fixed. I say **might** because we're not using `animateLayoutChanges` anywhere. 

It's happening on `main` branch as well, and in all cases when the snackbar is shown - I haven't figured out a solution...


a-szotyori (approved):
Thank you, Daniela!

PR #464, opened 2024-06-18
fix(bank-sdk): Fix Camera Permission flow for Example App

ndubkov-distcotech:
Ticket: PP-485


abolfazlimahdi (approved):
Looks great. Thank you :-)

PR #465, opened 2024-06-19
fix(capture-sdk): Fix page_swiped Analytics event

ndubkov-distcotech:
Ticket: PP-545


abolfazlimahdi (approved):
It looks good. Thank you!

PR #466, opened 2024-06-20
feat(bank-sdk): Redesign configuration screen in example app

abolfazlimahdi:
PP-516

There is no switch added to enable/disable the user tracking in the SDK. Only configuration screen is redesigned!


bank-sdk/example-app/src/main/java/net/gini/android/bank/sdk/exampleapp/ui/ConfigurationActivity.kt
```diff
@@ -32,7 +32,7 @@ class ConfigurationActivity : AppCompatActivity() {
         super.onCreate(savedInstanceState)
         binding = ActivityConfigurationBinding.inflate(layoutInflater)
         setContentView(binding.root)
-
+        configurationViewModel
```
ndubkov-distcotech, line 35: [Request] Looks like this one is not used

ndubkov-distcotech (approved):

PR #467, opened 2024-06-20
feat(health-api-library): Compresses images larger than 10MB to a JPEG with quality of 50

a-szotyori:
Using the same quality setting as in the Capture SDK.

IPC-276


danicretu (approved):
Looks good!

PR #469, opened 2024-06-24
ci(capture-sdk): Run instrumented tests in the check workflow

a-szotyori:
Should be merged after the scheduled releases have been published.


abolfazlimahdi (approved):
Looks great. Thank you!

PR #474, opened 2024-07-01
fix(health-sdk): Fix `Share With` flow broken when clients pop backst…

danicretu:
…ack after getting payment request callback

- Added a broadcast receiver to intercept taps on the system share sheet
- Move sending of `OpenBankState.Success` to after user picks an external app to share to
- Drawback is that the `OpenBankState.Success` event is not emitted if the user dismisses the share sheet - but then I guess it's not of interest to the client to know that the `paymentRequest` was created, but not finalised

IPC-290

### DIFF
--- a/health-sdk/sdk/src/main/java/net/gini/android/health/sdk/review/ReviewFragment.kt
+++ b/health-sdk/sdk/src/main/java/net/gini/android/health/sdk/review/ReviewFragment.kt
@@ -626,7 +626,7 @@ class ReviewFragment private constructor(
     private fun createSharePendingIntent() =  PendingIntent.getBroadcast(
         requireContext(), CHOOSER_REQUEST_ID,
         Intent(requireContext(), ShareWithBroadcastReceiver::class.java),
-        PendingIntent.FLAG_MUTABLE or PendingIntent.FLAG_UPDATE_CURRENT
+        PendingIntent.FLAG_IMMUTABLE
     )
 
     override fun onSaveInstanceState(outState: Bundle) {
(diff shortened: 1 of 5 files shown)


health-sdk/sdk/src/main/java/net/gini/android/health/sdk/review/ReviewFragment.kt
```diff
@@ -609,13 +623,28 @@ class ReviewFragment private constructor(
         }
     }
 
+    private fun createSharePendingIntent() =  PendingIntent.getBroadcast(
+        requireContext(), CHOOSER_REQUEST_ID,
+        Intent(requireContext(), ShareWithBroadcastReceiver::class.java),
+        PendingIntent.FLAG_MUTABLE or PendingIntent.FLAG_UPDATE_CURRENT
```
a-szotyori, line 629: [Suggestion] According to the [documentation](https://developer.android.com/reference/android/app/PendingIntent#FLAG_MUTABLE) _"It is strongly recommended to use [FLAG_IMMUTABLE](https://developer.android.com/reference/android/app/PendingIntent#FLAG_IMMUTABLE) when creating a PendingIntent."_. Is there something forcing us to use `FLAG_MUTABLE`?

a-szotyori (commented):
Nice work, I just left a suggestion and pushed a small fix because the `openBankState` flow was not reset to `PaymentState.NoAction` when the `ReviewFragmant` was dismissed (its `ReviewViewModel`'s scope was cancelled which lead to also cancelling the coroutine that resets the `openBankState`).

health-sdk/sdk/src/main/java/net/gini/android/health/sdk/review/ReviewFragment.kt
danicretu: Nope, there's no need for it to be mutable. Updated in the latest commit.

a-szotyori (approved):
Thank you! Can go to QA 🎉

PR #475, opened 2024-07-02
Ec 65 show install app and share with onboarding bottom sheets when review screen is not shown

danicretu:
Added `OpenWith` and `InstallApp` bottom sheets to `MerchantSDK`


merchant-sdk/sdk/src/main/java/net/gini/android/merchant/sdk/GiniMerchant.kt
```diff
@@ -125,7 +125,7 @@ class GiniMerchant(
      * successful payment requests or which screen is being displayed.
      */
 
-    private val _eventsFlow: MutableSharedFlow<MerchantSDKEvents> = MutableSharedFlow(extraBufferCapacity = 1)
+    private val _eventsFlow: MutableSharedFlow<MerchantSDKEvents> = MutableSharedFlow(extraBufferCapacity = 1,replay = 1)
```
danicretu, line 128: Added `replay` for this `sharedFlow` for the following use case:
- `ShareSheet` shown -> tap on an application -> tap event emitted -> the tap is received in the `viewModel` -> would be received in `onResume` of the fragment.

Since it's a `sharedFlow`, this value will be lost before `onResume` would be called. Adding `replay` will emit the last value when receivers register.

a-szotyori (approved):
Great work and thanks for the fun discussions regarding the kotlin flows! 👏

PR #477, opened 2024-07-03
feat(health-sdk):Set "powered by Gini" removable from payment component

llevente:


a-szotyori (changes requested):
Thank you, Lenci! Looks great I have only two requests:
1.  Please change the git commit message to match our [git commit template](https://github.com/gini/gini-mobile-android/blob/main/.git-stuff/commit-msg-template.txt) (it's ok to use rebase in feature branches to rewrite the history):
```
feat(health-sdk): Set "powered by Gini" removable from payment component

IPC-293
```
2. Please add a unit test for this. You can use Robolectric, if needed.

Thank you!

danicretu (commented):
Looks great, thank you! The only comment I have is the same as Alpár's about the git commit message.

a-szotyori (approved):
Thank you for adding the test! I approved the PR since the code is good and it can go to QA.

llevente:
@a-szotyori  @danicretu  Commit message is fixed, test added for checking if the view if hidden with this option.

llevente:
@a-szotyori  @danicretu  Commit message is also fixed, so the rule is kept. Moving forward to QA as Alpar said in the previous comment

PR #481, opened 2024-07-03
refactor(merchant-sdk): Validate `amount` field on the go, instead of…

danicretu:
Always show amount validation error, instead of hiding it if the amount field is focused -> when it's the only editable field, it will always be focused.


a-szotyori (approved):
Thank you!

PR #483, opened 2024-07-04
refactor(health-sdk): Use `async` in example app to parallelize test invoice uploads and refresh requests

a-szotyori:
I applied a learning from a coroutines talk to use `map { async {} }` to make test invoice uploads and refreshes faster.


danicretu (approved):
Really nice tidbit of knowledge! Thank you for sharing!

PR #489, opened 2024-07-10
feat(capture-sdk): Use new photo-picker.

GeraltRiv:
Now we are using new photo picker, for devices that doesn't have backdrop version, we use the same approach with list of apps PM-198


a-szotyori (changes requested):
Great work! I have a couple of minor comments. 

Please note that we are using comment labels to help prioritise the comments. You can read about them here: https://ginis.atlassian.net/wiki/spaces/PLMO/pages/83722835/Pull+Request+labels+description

a-szotyori:
I updated the target branch to collect this in the `PM-86-android-15` branch so that the Photopayment team can release their features. After QA we can then merge this PR into `PM-86-android-15`. In August we then merge `PM-86-android-15` and do a maintenance release.

PR #491, opened 2024-07-11
Ipc 295 use payment state

llevente:
[Comment]: One test is failing on the CI (the main one which is checking if state = Payable), but for me in the Android Studio and also in the example app, everything is ok. I'm not sure if my code is wrong, maybe the test document?!


health-sdk/sdk/src/main/java/net/gini/android/health/sdk/GiniHealth.kt
```diff
@@ -128,7 +128,7 @@ class GiniHealth(
 
     /**
      * Checks whether the document is payable by fetching the document and its extractions from the
-     * Gini Pay API and verifying that the extractions contain an IBAN.
+     * Gini Pay API and verifying that the extraction's payment state is "Payable".
```
a-szotyori, line 131: [Comment] Thank you for updating the documentation!

a-szotyori (approved):
Thank you! Can go to QA! 🎉

PR #492, opened 2024-07-11
feat(merchant-sdk): Add Orders Screen and Order Details Screen to the example app

a-szotyori:
Added five hardcoded order bank transfer details. These can be edited on the Order Details Screen.

Also the Orders Screen allows to launch the Order Details Screen for entering custom bank transfer details.

EC-74


danicretu (approved):
Thank you for handling this @a-szotyori ! Looks good, I will merge it into EC-74 and continue integrating it there.

PR #493, opened 2024-07-12
fix(capture-sdk): Fix Handler delayed call

GeraltRiv:
We need to check if fragment visible, if not we don't need to show snackBar, it cause a crash
PM-91


danicretu (approved):
Looks good, thank you!

PR #495, opened 2024-07-18
Ipc 283 custom localization manager

llevente:


a-szotyori (changes requested):
Thank you, Lenci, the code looks good! Unfortunately I found that some texts are not localized to the selected language.

Can you please also merge the target branch into this branch to fix the merge conflicts.

Also please fix the failing tests.

I used an emulator with Android 14 (API Level 34) and set the device language to English and the SDK language to German:

### Payment Component
The text on the button was not in German:
Screenshot_20240718_180330

### More Information (FAQ) Screen
Some of the texts were in German and some were in English. All of them should be in German:
Screenshot_20240718_180349

### Payment Review Screen
The text on the button was not in German:
Screenshot_20240718_180405

### "Install App" Bottom Sheet
None of the text is in German:
Screenshot_20240718_180501
Screenshot_20240718_180525

### "Share with" Onboarding Bottom Sheet
None of the text is in German:
Screenshot_20240718_180541

Thank you!

danicretu (changes requested):
Great work, thank you! I added some requests that I believe would help fix some of Alpar's comments also.

Also, please add the ticket number in the commit message, as specified in the [git commit template](https://github.com/gini/gini-mobile-android/blob/main/.git-stuff/commit-msg-template.txt).

health-sdk/sdk/src/main/java/net/gini/android/health/sdk/paymentcomponent/PaymentComponentView.kt
```diff
@@ -87,7 +89,7 @@ class PaymentComponentView(context: Context, attrs: AttributeSet?) : ConstraintL
             launchJobs()
         }
 
-    private val binding = GhsViewPaymentComponentBinding.inflate(getLayoutInflaterWithGiniHealthTheme(), this)
+    private var binding = GhsViewPaymentComponentBinding.inflate(getLayoutInflaterWithGiniHealthThemeAndLocale(GiniHealth.getSDKLanguage(context)?.languageLocale()), this)
```
GeraltRiv, line 92: @llevente 
I think here must be val, we don't need to reassign binding.

health-sdk/sdk/src/main/java/net/gini/android/health/sdk/paymentcomponent/PaymentComponent.kt
```diff
@@ -23,7 +23,7 @@ import net.gini.android.health.sdk.paymentprovider.getPaymentProviderApps
  *
  * It requires a [GiniHealth] instance and a [Context] (application or activity) to be created.
  */
-class PaymentComponent(private val context: Context, private val giniHealth: GiniHealth) {
+class PaymentComponent(private val context: Context, internal val giniHealth: GiniHealth, private var configuration: PaymentComponentConfiguration = PaymentComponentConfiguration()) {
```
GeraltRiv, line 26: @llevente 
Is it mistake, or we changed something that we need to use internal against private?

llevente: @GeraltRiv  we have to get the language from GiniHealth through the payment component so it has to be accessable. Initially i made a getter for the private field, but Alpar suggested it can be internal, so the module could see it.

GeraltRiv (approved):

a-szotyori (approved):
Thank you! 👏

PR #499, opened 2024-07-19
Ec 74 implement support for payment initiation without invoice document

danicretu:


merchant-sdk/sdk/src/main/java/net/gini/android/merchant/sdk/GiniMerchant.kt
```diff
@@ -0,0 +1,326 @@
...
+        }
+    }
+
+    /**
+     * Sets a lifecycle observer to handle state restoration after the system kills the app.
+     *
+     * @param registryOwner The SavedStateRegistryOwner to which the observer is attached.
+     * @param retryScope Should be a scope [setDocumentForReview] would be called in (ex: viewModelScope).
+     */
+    fun setSavedStateRegistryOwner(registryOwner: SavedStateRegistryOwner, retryScope: CoroutineScope) {
```
danicretu, line 234: @a-szotyori , not sure if this function is still needed? It's not used anywhere...
If it's needed I can update it to use `paymentDetails` instead of `document`... but not sure if that's the case?
If not, then I can cleanup this class even more.

a-szotyori (approved):
Thank you Daniela and Lenci for all the effort you put into this PR! 🙇

a-szotyori (approved):

PR #500, opened 2024-07-22
Ipc 293 removable poweredby

llevente:
This is just the fix of the example app, after Upload on Review the payment component was not getting the configuration for hiding the "powered by...". The logic is the same


danicretu (approved):
Looks good, thank you!

PR #508, opened 2024-07-25
ci(bank-sdk): Enable detekt for BankSdk and CaptureSdk

ndubkov-distcotech:
Proposition to enable detekt for modules:
`:capture-sdk:sdk`
`:bank-sdk:sdk`
`:bank-sdk:example-app`

Another modules will be untouched. 
Also, the baseline files are generated for this modules and prevent CI from failing based on already exist issues. **Already exist issues will be skipped and only newly created issues will be notified by detekt**


a-szotyori (approved):
Thank you, Nick!

PR #512, opened 2024-07-30
fix(capture-sdk): Fix a crash with navigation.

GeraltRiv:
If we close fragment before photo recognition finished, it cause a crash. This happens because we try to navigate from wrong location. That's why I have added safeNavigation function PP-496

### DIFF
--- a/capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/ui/FragmentImplCallback.java
+++ b/capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/ui/FragmentImplCallback.java
@@ -1,6 +1,5 @@
 package net.gini.android.capture.internal.ui;
 
-import android.app.Activity;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.view.View;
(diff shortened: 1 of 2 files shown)


capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/ui/FragmentImplCallback.java
```diff
@@ -54,4 +57,13 @@ void showAlertDialog(@NonNull final String message,
     @NonNull
     NavController findNavController();
 
+    @NonNull
+    default void safeNavigate(@NonNull final NavDirections navDirections) {
+        try {
+            findNavController().navigate(navDirections);
+        } catch (Exception exception) {
+            Logger logger = LoggerFactory.getLogger(FragmentImplCallback.class);
+            logger.error("Navigation exception " + exception.getMessage());
```
llevente, line 66: [Suggestion]: If this logger does not show a UI message shouldn't we show something to the user, why the flow is blocked and why the next step couldn't be reached?

llevente (approved):

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/ui/FragmentImplCallback.java
GeraltRiv: Yes, we can face a crash here if fragment already closed.
So I think it's not good to use this function for all navigation cases, just for this risky ones

PR #516, opened 2024-08-05
fix(mobile-android) Change navigationBar color

GeraltRiv:
PP-196

### DIFF
--- a/bank-sdk/example-app/src/main/res/values/styles.xml
+++ b/bank-sdk/example-app/src/main/res/values/styles.xml
@@ -10,6 +10,7 @@
         <item name="dividerColor">@color/gc_color_02</item>
         <item name="colorSurface">@color/gc_color_01</item>
         <item name="colorOnSurface">?attr/colorOnBackground</item>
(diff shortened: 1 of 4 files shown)


GeraltRiv:
> Sorry for being a [PITA](https://en.wiktionary.org/wiki/PITA#English) about commits 🙈 Please rewrite the commit message (it's ok to force push into your own feature branch): we don't use `mobile-android` as a project name. When multiple projects are affected, then the project has to be omitted. I usually list in the commit body the affected projects. The commit message should then look like this:
> 
> ```
> fix: Change navigationBar color
> 
> Affects: bank-sdk, capture-sdk, health-sdk
> 
> PP-196
> ```
> 
> For the future: what we prefer is to have separate commits in such cases for each modified project to have a dedicated `fix(<project>)` commit for each one.
> 
> The advantage of having the project name in the commit is that in our monorepo we can filter the git history to see what has been changed for each project, for ex. to view all changes for Bank SDK since a particular release: `$ git log --oneline --grep="bank-sdk" "bank-sdk;3.12.0..HEAD" --no-merges`
> 
>  git-log.mov

Done

PR #518, opened 2024-08-05
ci(bank-sdk): Update detekt and adjust for Compose

ndubkov-distcotech:


a-szotyori (approved):
Thank you, Nick, for helping us make and keep our code cleaner! 👏

PR #519, opened 2024-08-06
Ec 68 health api version config

llevente:
Please check if documentation should be updated about the new apiVersion param for GiniHealthAPIBuilder


a-szotyori (approved):
Thank you, Lenci!

PR #525, opened 2024-08-07
fix(capture-sdk) Add contentDescription for tablet layouts

GeraltRiv:
PP-688


a-szotyori (approved):
Thank you, Eugene! As these changes are so simple I think it's safe to make an exception and skip the dedicated QA for this and merge it directly into `main`.

PR #531, opened 2024-08-12
Ec 117 update payment review screen according to design

danicretu:
Please check commit details for further info


a-szotyori (approved):
Looks great! I did a small fix so that the edited order details are used to start the payment flow. Let me know please, if it's ok how I did it.

PR #540, opened 2024-08-20
feat(health-sdk): Error message: Multiple documents in a claim

danicretu:
IPC-307

### DIFF
--- a/health-sdk/sdk/src/main/java/net/gini/android/health/sdk/GiniHealth.kt
+++ b/health-sdk/sdk/src/main/java/net/gini/android/health/sdk/GiniHealth.kt
@@ -145,7 +145,6 @@ class GiniHealth(
      * Gini Pay API and verifying that the extraction's payment state is "Payable".
      *
      * @return `true` if the document is payable and `false` otherwise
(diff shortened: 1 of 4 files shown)


health-sdk/sdk/src/main/java/net/gini/android/health/sdk/GiniHealth.kt
```diff
@@ -145,17 +145,43 @@ class GiniHealth(
      * Gini Pay API and verifying that the extraction's payment state is "Payable".
      *
      * @return `true` if the document is payable and `false` otherwise
-     * @throws Exception if there was an error while retrieving the document or the extractions
```
a-szotyori, line 148: [Request] We need to keep the @throws here and also add it to `checkIfDocumentContainsMultipleDocuments()` because `getExtractionsForDocument()` is private and so clients won't know that `checkIfDocumentIsPayable()` and `checkIfDocumentContainsMultipleDocuments()` can throw exceptions.

a-szotyori (changes requested):
Looks great! Just one small reference documentation related request.

a-szotyori (approved):

a-szotyori (approved):

PR #541, opened 2024-08-20
feature(bank-sdk) Add possibility to input client credentials 

GeraltRiv:
Now in settings page we can change clientId and clientSecret PP-712

### DIFF
--- a/bank-sdk/example-app/src/main/java/net/gini/android/bank/sdk/exampleapp/core/DefaultServiceNetworkApi.kt
+++ b/bank-sdk/example-app/src/main/java/net/gini/android/bank/sdk/exampleapp/core/DefaultServiceNetworkApi.kt
@@ -1,0 +1,90 @@
+package net.gini.android.bank.sdk.exampleapp.core
+
+import android.content.Context
+import android.text.TextUtils
+import net.gini.android.bank.api.GiniBankAPI
+import net.gini.android.bank.api.GiniBankAPIBuilder
+import net.gini.android.bank.sdk.exampleapp.R
+import net.gini.android.capture.network.GiniCaptureDefaultNetworkService
+import net.gini.android.core.api.DocumentMetadata
+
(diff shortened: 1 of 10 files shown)


bank-sdk/example-app/src/main/java/net/gini/android/bank/sdk/exampleapp/ui/ConfigurationActivity.kt
```diff
@@ -616,4 +651,9 @@ class ConfigurationActivity : AppCompatActivity() {
         }
     }
 
+    private fun checkClientSecretAndClientId() {
```
a-szotyori, line 654: [Suggestion] I think if we rename the method to `applyClientSecretAndClientId()` it reflects more accurately what it does.

bank-sdk/example-app/src/main/java/net/gini/android/bank/sdk/exampleapp/core/DefaultServiceNetworkApi.kt
```diff
@@ -0,0 +1,90 @@
...
+
+import android.content.Context
+import android.text.TextUtils
+import net.gini.android.bank.api.GiniBankAPI
+import net.gini.android.bank.api.GiniBankAPIBuilder
+import net.gini.android.bank.sdk.exampleapp.R
+import net.gini.android.capture.network.GiniCaptureDefaultNetworkService
+import net.gini.android.core.api.DocumentMetadata
+
+class DefaultServiceNetworkApi (internal val context: Context, internal val logger: org.slf4j.Logger) {
```
a-szotyori, line 11: [Suggestion] I think the name `DefaultNetworkServicesProvider` what it does better.

a-szotyori (commented):
Looks great! I only have naming suggestions (because [naming is hard](https://martinfowler.com/bliki/TwoHardThings.html) 😬.

a-szotyori (approved):
Thank you! Can go to QA 🎉

PR #543, opened 2024-08-21
ci: Updated github action versions, removed unused workflows

llevente:
PM-99

@a-szotyori  Saw that there is a branch where you deleted the unused workflows. I deleted them in this branch also so we avoid re adding them if the merging order will change.


a-szotyori:
Thank you, @llevente, I removed my PR.

a-szotyori (approved):
Thank you, Lenci! I will go ahead and merge it. Unfortunately currently there is no other way to see if there are issues.

PR #544, opened 2024-08-21
ci: Update maven package developer info

a-szotyori:
The time has come to pass the torch to you Mahdi, or as Gemini would put it in Lord of the Rings style:
"Take this banner, Mahdi. It is heavy with the hopes of many, and the weight of Gini. May your hand be steady as you carry it forward. May your heart be filled with wisdom and justice. For you, Mahdi, the torch has been lit. Let it burn bright."


abolfazlimahdi (approved):
Looks great. Thank you 🥇

PR #549, opened 2024-08-26
Bac 1531 spike for adding data to http headers on mobile side

danicretu:


capture-sdk/sdk/src/main/java/net/gini/android/capture/util/UploadMetadata.kt
```diff
@@ -0,0 +1,84 @@
+package net.gini.android.capture.util
```
a-szotyori, line 1: [Request] `UploadMetadata` is closely coupled to the documents and so I think it would fit better into the `net.gini.android.capture.document` package.

a-szotyori (changes requested):
Great work! 👍 I only have minor requests/suggestions.

a-szotyori (approved):
Awesome! Please upload a build to App Center and share it in the ticket. I will inform backend that they can use it for sending upload metadata headers to the Pay API. Once they confirm that it works we can then move BAC-1531 to done and move on with ticket ~[BAC-894](https://ginis.atlassian.net/browse/BAC-894)~ [BAC-893](https://ginis.atlassian.net/browse/BAC-893) to finalize the feature.

What do you think of creating a branch for ~BAC-894~ BAC-893 and setting that one as the target branch for this PR? Then when work is continued with ~BAC-894~  BAC-893 we have a branch for it and once we are done we can merge it into "main".

PR #556, opened 2024-08-29
feature(capture-sdk) Add gini payment QR

GeraltRiv:
Added possibility to detect https://payment.gini.net/* qr codes. This feature can't be disabled EC-46

### DIFF
--- a/capture-sdk/sdk/src/main/java/net/gini/android/capture/GiniCaptureFragment.kt
+++ b/capture-sdk/sdk/src/main/java/net/gini/android/capture/GiniCaptureFragment.kt
@@ -8,6 +8,9 @@ import androidx.fragment.app.Fragment
 import androidx.fragment.app.FragmentFactory
 import androidx.navigation.NavController
 import androidx.navigation.fragment.findNavController
+import com.google.mlkit.vision.barcode.BarcodeScannerOptions
+import com.google.mlkit.vision.barcode.BarcodeScanning
+import com.google.mlkit.vision.barcode.common.Barcode
 import net.gini.android.capture.analysis.AnalysisFragment
 import net.gini.android.capture.analysis.AnalysisFragmentDirections
 import net.gini.android.capture.analysis.AnalysisFragmentListener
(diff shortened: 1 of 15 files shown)


capture-sdk/sdk/src/main/java/net/gini/android/capture/GiniCaptureFragment.kt
```diff
@@ -294,10 +297,15 @@ class CaptureFragmentFactory(
...
             CameraFragment::class.java.name -> return CameraFragment().apply {
+                val options = BarcodeScannerOptions.Builder()
+                    .setBarcodeFormats(Barcode.FORMAT_QR_CODE)
+                    .build()
+
                 setListener(
                     cameraListener
                 )
                 setCancelListener(cancelListener)
+                setBarcodeScanner(BarcodeScanning.getClient(options))
```
a-szotyori, line 308: [Clarification] Why do we need to pass the barcode scanner here?

Previously we encapsulated the MLKit dependency in `QRCodeDetectorTaskMLKit()` so that we can swap it out with some other QR code detector in the future.

I would prefer if we could avoid exposing MLKit BarcodeScanner.

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/qrcode/GiniPaymentParser.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Internal use only.
+ * QR Code parser for the eps payment.gini QR Code url.
+ *
+ * @suppress
+ */
+class GiniPaymentParser implements QRCodeParser<PaymentQRCodeData> {
+
+    private static final String GINIPAYMENT_HOTS = "payment.gini.net";
```
a-szotyori, line 15: [Request] Typo: `GINIPAYMENT_HOST`

a-szotyori (commented):
Thanks, looks good and I just have a clarification questions (and saw a small typo).

capture-sdk/sdk/src/main/java/net/gini/android/capture/GiniCaptureFragment.kt
a-szotyori: I found a way to avoid exposing MLKit by making the `CameraFragmentImpl.initQRCodeReader()` method protected and then overriding it in a subclass in the tests to do nothing.

Let me know please if that is ok for you.

a-szotyori (approved):
Thank you! 👍

PR #561, opened 2024-09-04
fix(core-api-library): Convert headers to lowercase

a-szotyori:
This fixes the `LocationResponse` parser in the Health API Library.

_Edited again to relaunch workflows._


abolfazlimahdi (approved):
Looks good. Thank you!

PR #567, opened 2024-09-10
Run health api library check manually on stage

a-szotyori:
Allows running the Health API integration tests also on staging (`health-api.stage.gini.net` and `user.stage.gini.net`).

There are a few differences when running tests on staging:
- certificate pinning is disabled because staging hosts use self-signed SSL certificates and pinning fails due to the lack of a trust anchor (`java.security.cert.CertPathValidatorException: Trust anchor for certification path not found.`)
- running tests on Android 23 or older is disabled for the same reason as above
- `testGetPayment()` is disabled because it needs access to the staging Pay API (pay-api.pia.stage.gini.net) which requires VPN access (via Tailscale).

Workflow can be manually triggered from [here](https://github.com/gini/gini-mobile-android/actions/workflows/health-api-library.check.yml) by using the branch `run-health-api-library-check-manually-on-stage` (after PR is merged the `main` branch will be needed) and selecting “production” or “staging” from the dropdown:
Screenshot 2024-09-10 at 10 42 12

### DIFF
--- a/.github/workflows/health-api-library.check.yml
+++ b/.github/workflows/health-api-library.check.yml
@@ -123,11 +123,11 @@ jobs:
           script: >
             adb uninstall net.gini.android.health.api.test ;
             ./gradlew -Pandroid.testInstrumentationRunnerArguments.apiEnv=${{ inputs.healthApiEnvironment }} health-api-library:library:connectedCheck
-            -PtestClientId="gini-mobile-test"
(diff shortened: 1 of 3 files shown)


zladzeyka (approved):
Great job @a-szotyori 👏🏻

abolfazlimahdi (approved):
Looks great. Thank you 🥇

PR #570, opened 2024-09-11
Ec 142 setup maven central publishing

a-szotyori:
Besides the release workflow I also added a check workflow. I saw the Internal Payment SDK has unit tests.

Important: the workflows will either not trigger or they will fail, because it assumes `gini-internal-payment-sdk/library` will be renamed to `internal-payment-sdk/sdk`. We can ignore failures until the renaming happens.


danicretu (approved):
Looks good! Thank you for adding the CI flows 🙏🏻

PR #572, opened 2024-09-11
fix: (capture-sdk) Ignore CancellationException

GeraltRiv:
Ignore CancellationExcpetion because our qrcode detection task was just canceled, we don't need to trigger onQRCodeReaderFail event PP-814


a-szotyori (approved):
Thank you for the quick fix!

PR #575, opened 2024-09-13
fix(capture-sdk): Remove button in description

GeraltRiv:
Left only name of the button, without "button" word, because talkback pronounce button twice PP-220


abolfazlimahdi (approved):
Looks good. Thank you!

PR #576, opened 2024-09-13
fix(capture-sdk): Change gini_payment host

GeraltRiv:
EC-46


abolfazlimahdi (approved):
Looks good. Thank you!

PR #577, opened 2024-09-16
Pp 548 increase buttons size

GeraltRiv:


danicretu (approved):
Looks good, thank you!

PR #579, opened 2024-09-17
Ec 46 iban fix

GeraltRiv:
@mahdiabolfazli 
Somewhere we trying to close image after detection twice, this cause IllegalStateException
I propose to check this exception, but after release research and found problem why we try to close image twice.


abolfazlimahdi (approved):
Thanks looks good. I'll ask our QA engineer to create a bug ticket for this (closing the image twice).

PR #588, opened 2024-09-26
Ipc 331 doctor name extraction

llevente:
Made extractions public in PaymentDetails so clients can reach the specific extractions during review and use the data as they want. No new UI element on review component.


GeraltRiv (approved):
Looks good

abolfazlimahdi (approved):
Thank you!

PR #590, opened 2024-10-03
fix(capture-sdk): Make sure that frame will not be closed before all …

GeraltRiv:
…image detection finished

PP-823


capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragmentImpl.java
```diff
@@ -1881,8 +1886,6 @@ public void onPreviewFrame(@NonNull Image image, @NonNull Size imageSize, int ro
                 }
 
                 if (mPaymentQRCodeReader != null) {
-                    previewFrameReferenceCount.getAndIncrement();
```
GeraltRiv, line 1884: Sometimes async thread inside `ibanRecognizerFilter.processImage` finish detection fast enough, that our 
`if (mPaymentQRCodeReader != null) ` not called, so we  are closing image before paymentQrCodeReader started his work. So I firstly increment counter, than starting async threads
@mahdiabolfazli 
@llevente

llevente (approved):

abolfazlimahdi (approved):
Looks good. Thank you!

PR #599, opened 2024-10-17
Ipc 327 rebuild the health sdk to use the basis sdk removing all duplications

llevente:
This is the finalization of the whole Basis SDK integration into Merchant and Health.


internal-payment-sdk/sdk/src/main/java/net/gini/android/internal/payment/review/reviewComponent/ReviewView.kt
```diff
@@ -217,17 +238,34 @@ class ReviewView(private val context: Context, attrs: AttributeSet?) :
...
+            setDisabledIcon(context.getString(R.string.gps_purpose_hint), binding.purposeLayout)
+            binding.purpose.isEnabled = false
+        }
     }
 
     private fun handleInputFocusChange(hasFocus: Boolean, textInputLayout: TextInputLayout) {
         if (hasFocus) textInputLayout.hideErrorMessage() else textInputLayout.showErrorMessage()
     }
 
+//    fun clearFocus() {
```
GeraltRiv, line 259: Commented code, do we need to delete it now?

GeraltRiv (commented):
Everything good. Added some small suggestions

internal-payment-sdk/sdk/src/main/java/net/gini/android/internal/payment/review/reviewComponent/ReviewView.kt
danicretu: Thanks for catching that - deleted it.

GeraltRiv (approved):

PR #601, opened 2024-10-19
feature(bank-sdk): Added double tap to zoom feature on Invoice Preview screen

ndubkov-distcotech:
PP-884


abolfazlimahdi (commented):
[Request] The double tap functionality should be the same as when we are tapping on a picture in Preview screen. On the first double tap it should zoom in and on the second one it should zoom out. Thank you!

abolfazlimahdi (approved):
Looks good. Thank you!

PR #609, opened 2024-10-24
feat(basis-sdk): Chaged payment status invalid to paid_adjusted

llevente:
@mahdiabolfazli  FYI: Core api is changed in this matter, but only the enum value naming. Let me now if this is ok.


GeraltRiv (approved):
All good. Thank you

PR #615, opened 2024-11-05
fix: (bank-sdk) Add safe navigation and increase interval

GeraltRiv:
https://ginis.atlassian.net/browse/PP-821

If you tap on button really fast sometimes it crashes

Also found a problem, if you have a lot of elements in DigitalInvoiceFragment list you can tap really fast on 2 edit buttons of different elements, in this case `IntervalClickListener` will not help, so I added `safeNavigation`.


danicretu (approved):
Looks good, thank you!

PR #618, opened 2024-11-12
feature(bank-sdk): Skonto + RA validation implementation

ndubkov-distcotech:
PP-795


abolfazlimahdi (changes requested):
Code looks fine. The behavior is not as it was intended. The validation error show for a moment and then disappears. It should disappear with a user's action. Thank you!

abolfazlimahdi (approved):
Looks good. Thank you!

PR #625, opened 2024-11-25
fix(basis-sdk): Open with bottom sheet layout fix

llevente:
IPC-438


danicretu (approved):
Looks good, thank you!

PR #628, opened 2024-11-25
PP 835 updating libraries

obaidgini:


abolfazlimahdi (changes requested):
Looks good. Thank you! Only two points:

1. Can you please refactor the libs.versions file and extract the libraries that are added without separate versions like: `androidx-core-ktx = "androidx.core:core-ktx:1.14.0-alpha01"`
2. It would be great to have a look at changes that come with major versions of Kotlin and AGP. We also want to notify our clients about upgrading these two libraries in the future.

abolfazlimahdi (approved):
Looks good. Thank you!

PR #632, opened 2024-11-28
Ipc 490 update docs  and code

llevente:


health-sdk/sdk/src/doc/source/event-tracking.rst
```diff
@@ -24,6 +24,25 @@ The ``GiniHealth`` class exposes kotlin flows which you can collect to track eve
...
    * ``PaymentState.Cancel()`` when the user cancelled the payment flow.
+   .. warning::
+       As the SDK is not responsible for navigation flows outside of it, removing the payment fragment from the hierarchy is the responsibility of implementers at ``PaymentState.Success(paymentRequest)`` or ``PaymentState.Cancel()`` events.
+.. code-block:: java
+            launch {
+                    viewModel.openBankState.collect { paymentState ->
+                        when (paymentState) {
+                            is GiniHealth.PaymentState.Success -> {
+                                ...
+                                supportFragmentManager.popBackStack()
```
danicretu, line 35: [Request] Please replace this with a comment like // Remove fragment, since we do not know if the clients will be using classic or NavGraph in their applications.

health-sdk/sdk/src/doc/source/event-tracking.rst
```diff
@@ -24,6 +24,25 @@ The ``GiniHealth`` class exposes kotlin flows which you can collect to track eve
...
+
+                            }
+                            is GiniHealth.PaymentState.Cancel -> {
+                                supportFragmentManager.popBackStack()
```
danicretu, line 39: [Request] Same comment as above

health-sdk/sdk/src/doc/source/event-tracking.rst
```diff
@@ -24,6 +24,25 @@ The ``GiniHealth`` class exposes kotlin flows which you can collect to track eve
    * ``PaymentState.Success(paymentRequest)`` when the payment request is ready and the banking app will be opened.
    * ``PaymentState.Error(throwable)`` when there was an error creating the payment request or opening the banking app.
    * ``PaymentState.Cancel()`` when the user cancelled the payment flow.
+   .. warning::
+       As the SDK is not responsible for navigation flows outside of it, removing the payment fragment from the hierarchy is the responsibility of implementers at ``PaymentState.Success(paymentRequest)`` or ``PaymentState.Cancel()`` events.
+.. code-block:: java
+            launch {
+                    viewModel.openBankState.collect { paymentState ->
```
danicretu, line 31: [Suggestion] Maybe it would be clearer if we used `giniHealth.openBankState` ? What do you think?

danicretu (changes requested):
Looks good! Could you please add the same warning also in `flow.rst` after the `Create the PaymentFragment` ?

Just in case the developers only read the bare minimum of the documentation.

danicretu (approved):

PR #639, opened 2024-12-09
feat(health-sdk): Updated handling API response for non-payable docum…

llevente:
…ents

IPC-497


danicretu (approved):
Looks good, thank you!

PR #642, opened 2024-12-11
Ipc 407 alternative texts and payment flow fix

llevente:
This PR also contains fix for the payment flow for new users (skipping "Continue to overview" status after initial bank selection. After my own tests it is working as expected, but after review the Entry points for new user should go through a quick test.

### DIFF
--- a/health-sdk/sdk/src/main/java/net/gini/android/health/sdk/integratedFlow/PaymentFragment.kt
+++ b/health-sdk/sdk/src/main/java/net/gini/android/health/sdk/integratedFlow/PaymentFragment.kt
@@ -329,6 +329,14 @@ class PaymentFragment private constructor(
                     if (childFragmentManager.fragments.any { it is PaymentComponentBottomSheet }) {
                         return
                     }
+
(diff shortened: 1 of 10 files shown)


health-sdk/sdk/src/main/java/net/gini/android/health/sdk/integratedFlow/PaymentFragment.kt
```diff
@@ -329,6 +329,14 @@ class PaymentFragment private constructor(
                     if (childFragmentManager.fragments.any { it is PaymentComponentBottomSheet }) {
                         return
                     }
+
+                    if (!viewModel.giniInternalPaymentModule.getReturningUser() && viewModel.documentId != null) {
```
danicretu, line 333: [Clarification] Does this case only apply to flows with document? For flows without document, we have the same  steps as before?

danicretu (approved):

danicretu:
Not needed anymore - merged into https://github.com/gini/gini-mobile-android/tree/feat(health-sdk)-Accessibility

PR #646, opened 2024-12-18
Ipc 508 show ingredient only in payment component bank selection and review screen

danicretu:


GeraltRiv (approved):

obaidgini (approved):
Overall Looks good to me, Just a small point, which was discussed in the meeting that if the spelling of states for example FULL_VISIBLE are returned wrong by mistake (for example -> FUL_VISIBLE instead of FULL_VISIBLE) then app should not crash. It should just hide the logo. Please have a look and handle this case if it is not already handled. Thank you.

PR #655, opened 2025-01-20
IPC-532: SDK communication tone

obaidgini:
@danicretu , I have raised this PR to IPC-508 branch related to brand visibility, because this communication tone is also from the same API and you already parsed it. So, we can merge this into IPC-508 after passed review

### DIFF
--- a/internal-payment-sdk/sdk/src/main/java/net/gini/android/internal/payment/GiniInternalPaymentModule.kt
+++ b/internal-payment-sdk/sdk/src/main/java/net/gini/android/internal/payment/GiniInternalPaymentModule.kt
@@ -53,16 +53,11 @@ class GiniInternalPaymentModule(private val context: Context,
      * like if(tone == formal) getString(R.id.hello_formal) else getString(R.id.hello_informal).
      * */
 
-    internal fun updateSDKLanguageForInternalRecord(context: Context) {
+    private fun updateSDKLanguageForInternalRecord(context: Context) {
         if (!GiniPaymentPreferences(context).getLanguageOverriddenByUser()) {
             val sdkLanguage = when (Resources.getSystem().configuration.locales[0].language) {
-                "en" -> {
-                    GiniLocalization.ENGLISH
-                }
-
-                else -> {
-                    GiniLocalization.GERMAN
-                }
+                "en" -> GiniLocalization.ENGLISH
+                else -> GiniLocalization.GERMAN
             }
             GiniPaymentPreferences(context).saveSDKLanguage(sdkLanguage, false)
         }
@@ -157,6 +152,10 @@ class GiniInternalPaymentModule(private val context: Context,
                 is Resource.Success -> {
                     val ingredientBrandVisibility = IngredientBrandType.valueOf(configurations.data.ingredientBrandType)
                     saveIngredientBrandVisibility(ingredientBrandVisibility)
+
+                    configurations.data.communicationTone?.let { tone ->
+                        saveSDKCommunicationTone(tone.name)
+                    }
                 }
                 else -> {}
             }
@@ -275,7 +274,12 @@ class GiniInternalPaymentModule(private val context: Context,
         const val SHARE_WITH_INTENT_FILTER = "share_intent_filter"
 
         fun getSDKLanguage(context: Context): GiniLocalization? {
-            return GiniPaymentPreferences(context).getSDKLanguage()
+            val preferences = GiniPaymentPreferences(context)
+            val language = preferences.getSDKLanguage()
+            return when {
+                language == GiniLocalization.GERMAN && preferences.getSDKCommunicationTone() == CommunicationTone.INFORMAL -> GiniLocalization.GERMAN_INFORMAL
+                else -> language
+            }
         }
     }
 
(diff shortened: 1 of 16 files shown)


internal-payment-sdk/sdk/src/main/java/net/gini/android/internal/payment/GiniInternalPaymentModule.kt
```diff
@@ -37,8 +40,30 @@ class GiniInternalPaymentModule(private val context: Context,
         giniHealthAPI: GiniHealthAPI,
     ) : this(context) {
         _giniHealthAPI = giniHealthAPI
+        updateSDKLanguageForInternalRecord(context)
```
danicretu, line 43: [Request] This constructor is used when instantiating `GiniInternalPaymentModule` from `HealthSDK`.

Since this class is a dependency of both `HealthSDK` and `MerchantSDK`, could we also add a call to `updateSDKLanguageForInternalRecord` when using this class from `MerchantSDK` ?  The `MerchantSDK` uses the main constructor of `GiniInternalPaymentModule` to instantiate the class.

internal-payment-sdk/sdk/src/main/java/net/gini/android/internal/payment/GiniInternalPaymentModule.kt
```diff
@@ -200,20 +233,73 @@ class GiniInternalPaymentModule(private val context: Context,
...
+                apply()
+            }
+        }
+
+        fun getSDKCommunicationTone(): CommunicationTone =
+            sharedPreferences.getString(
+                SDK_COMMUNICATION_TONE_PREFS_KEY,
+                CommunicationTone.FORMAL.toString()
+            )?.let {
+                CommunicationTone.valueOf(
```
danicretu, line 249: [Request] Similar to the crash described [here](https://github.com/gini/gini-mobile-android/pull/646#pullrequestreview-2546441940) , the app will crash if backend sends an invalid value for `CommunicationTone`. 
Please consider handling this case, and returning a default value.

internal-payment-sdk/sdk/src/main/java/net/gini/android/internal/payment/GiniInternalPaymentModule.kt
```diff
@@ -200,20 +233,73 @@ class GiniInternalPaymentModule(private val context: Context,
                     it
                 )
             } ?: IngredientBrandType.INVISIBLE
+
+        fun saveSDKCommunicationTone(tone: String) {
```
danicretu, line 237: [Suggestion] I think this method can be `private`, since we only use it in this class.

danicretu (changes requested):
Thank you for this solution, it looks good! I have added just a couple of comments, most of them with some very small requests.
Please let me know if anything is unclear.

internal-payment-sdk/sdk/src/main/java/net/gini/android/internal/payment/GiniInternalPaymentModule.kt
obaidgini: @danicretu this method is used in unit test and also in the GiniPaymentInternalModule, same as saveIngredientBrandVisibility. If i will make it private it will not be accessible.

danicretu: @obaidgini , my mistake, I was referring to `saveSDKCommunicationTone` in `GiniInternalPaymentModule`, at line 194 🙈. I'm sorry for posting the comment at the wrong line, but was checking in the code and had the impression I was commenting on the other one.

obaidgini: Ah, no worries, thank you very much, that one is made private :)

internal-payment-sdk/sdk/src/main/java/net/gini/android/internal/payment/utils/GiniLocalizationInternal.kt
```diff
@@ -0,0 +1,54 @@
...
+ * Important Note!!
+ * As of now if user wants to override the strings, they have to put the same keys in default values
+ * folder.
+ * It will work as it is, but the customers who wants the support for Informal, and they want to
+ * override the informal strings, they need to make this folder in their project values-de and
+ * keep a string file in this folder and add the keys in this file (to override the informal strings),
+ * for the formal no change is made.
+ * */
+
+enum class GiniLocalizationInternal(val readableName: String) {
```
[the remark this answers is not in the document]
obaidgini: Sure thing, Just did not added by mistake , It was missed because I added it in the class name and missed to put in the start :D , will do now.

internal-payment-sdk/sdk/src/main/java/net/gini/android/internal/payment/GiniInternalPaymentModule.kt
obaidgini: introduced init block

danicretu (approved):
Thank you for the changes @obaidgini ! Looks great 👌

abolfazlimahdi (approved):
Looks good. Thank you!

PR #658, opened 2025-01-27
Ipc 556 reload payment providers if gini health sdk was instantiated without credentials or session manager

danicretu:
- Made `loadPaymentProviders` a public method again. Although when starting the payment fragment a call to this method is made, perhaps it will be safer to allow the developer to manually trigger this if any error occurs, or somehow their flow is different. Also useful because the `trustMarkers` flow depends on this call, and the trust markers may be needed before entering the payment flow.
- Removed exiting the payment flow if status of loading payment providers is `Error` or `Canceled`.
- Updated docs to mention `loadPaymentProviders` is a public method, and some context for when it should be manually triggered.

The `PaymentFragment` makes a call to `loadPaymentProviders` when it is started, so I think this will fix IBM's issue without them having to call the method manually.

@zladzeyka , I added you as well so double check if the updated info in the docs makes sense 🤔

### DIFF
--- a/health-sdk/sdk/src/doc/source/flow.rst
+++ b/health-sdk/sdk/src/doc/source/flow.rst
@@ -24,8 +24,12 @@ After that you can create an instance of ``GiniHealth``:
 
 .. note::
 
-    ``GiniHealth`` exposes a method for loading payment providers manually: ``giniHealth.loadPaymentProviders()``. It can be used if there was an error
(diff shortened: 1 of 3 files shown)


obaidgini (approved):
Look good to me, Thank you! @danicretu

zladzeyka (approved):
Looks good, thanks @danicretu

PR #662, opened 2025-02-04
docs(health-sdk): Add mention about method that checks whether a docu…

danicretu:
…ment contains multiple invoices

IPC-576

### DIFF
--- a/health-sdk/sdk/src/doc/source/flow.rst
+++ b/health-sdk/sdk/src/doc/source/flow.rst
@@ -104,6 +104,26 @@ Health API. You can then store the ``isPayable`` state in your own data model.
         }
     }
 
+Check if document has multiple invoices
+---------------------------------------
+
+Call ``giniHealth.checkIfDocumentContainsMultipleDocuments()`` with the composite document id to check whether it contains multiple invoices.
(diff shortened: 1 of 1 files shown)


health-sdk/sdk/src/doc/source/flow.rst
```diff
@@ -104,6 +104,26 @@ Health API. You can then store the ``isPayable`` state in your own data model.
         }
     }
 
+Check if document has multiple invoices
+---------------------------------------
+
+Call ``giniHealth.checkIfDocumentContainsMultipleDocuments()`` with the composite document id to check whether it contains multiple invoices.
```
obaidgini, line 110: Just a small comment, this sentence sounds incomplete because of whether, can we add "or not" in the end? 

with the composite document id to check whether it contains multiple invoices or not.

obaidgini (approved):
@danicretu I have added a small comment related to typo, Also approving the PR because your sentence is correct as well, I just added a suggestion, if that suggestion looks good to you then update it, otherwise it is also fine.
Thank you! :)

health-sdk/sdk/src/doc/source/flow.rst
danicretu: Thank you for your suggestion - implemented it.

PR #663, opened 2025-02-04
chore(CI): checking if the app tester link is also published by the C…

obaidgini:
…I to jira ticket or not

PM-139


zladzeyka (approved):
Thank you @obaidgini

PR #666, opened 2025-02-05
PP-947: Landscape orientation review screen

obaidgini:


capture-sdk/sdk/src/main/java/net/gini/android/capture/review/multipage/previews/MiddlePageManager.kt
```diff
@@ -48,14 +48,11 @@ class MiddlePageManager : LinearLayoutManager {
 
         if (lpPos != 0 && lpPos != itemCount - 1) return
 
-        //Post wait to obtain child width
-        //Pas the calculation to start and end padding to centre the child
-        child.post {
```
abolfazlimahdi, line 53: [Request] Are you sure it is safe to delete this code? Please double check. Thank you!

abolfazlimahdi (commented):
Looks good. I added one comment. Please take a look and verify it is intended. Thank you!

capture-sdk/sdk/src/main/java/net/gini/android/capture/review/multipage/previews/MiddlePageManager.kt
obaidgini: @mahdiabolfazli Yes, i have verified this code by testing the on emulator + real device.
Scenarios which I have tested
1- Current implementation does not have any impact. It is working fine on real device portrait mode and on emulator as well.
2- It is working fine for landscape mode.
3- It is working fine for landscape to portrait mode switch.

Reason to remove.
Because of this post method, the calculations were not done right away, Which results in wrong calculations when user was in landscape mode, specially when switching from portrait to landscape.
After debugging and removing post method, padding calculations run immediately, results in correct value, there was no impact seen while testing on real devices (s24 ultra + Galaxy A53) and emulator.

abolfazlimahdi (approved):
Looks good. Thank you! Please make sure Saranya will test that part of the code that you removed with extra care. Thanks :)

PR #667, opened 2025-02-11
Ipc 528 landscape in payment review screen change landscape to portrait mode document is not loaded

danicretu:


obaidgini (approved):
Looks good to me,   👌  Thank you!

PR #669, opened 2025-02-13
fix(internal-payment-sdk): Added `Locale.GERMAN` to amount formatter

danicretu:
IPC-566

### DIFF
--- a/health-sdk/example-app/src/main/java/net/gini/android/health/sdk/exampleapp/invoices/ui/model/InvoiceItem.kt
+++ b/health-sdk/example-app/src/main/java/net/gini/android/health/sdk/exampleapp/invoices/ui/model/InvoiceItem.kt
@@ -42,7 +42,7 @@ data class InvoiceItem(
                 val price = parsePrice(substrings[0])
                 val currency = Currency.getInstance(substrings[1])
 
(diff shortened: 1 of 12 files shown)


obaidgini (changes requested):
Hi @danicretu , Everything looks great, Just a small thing, formatCurrency will create a problem I think, because I used DecimalFormat("#,##0.00", symbols) format to format the amount in ReviewView. (IBM fix), I simulated it by commenting the amountWatcher in OrderDetailsFragment, and entering some values like 777.1 and creating a order, in the review screen I was receiving the amount and then sanitizing it if it was not a perfect two decimal place amount.

For you reference here are the method and classes which you can check
Util class -> CurrencyFormatter.kt, Method -> formatCurrency

danicretu:
Hi @obaidgini , I updated the PR, please have another look! 

Main changes:
- I removed the `isValidTwoDecimalNumber` method because using it was giving the wrong result in the following case: locale switched to German, and amount is `123.45` -> since in this case the decimal separator is `.`, the result was `12345.0`, because it bypassed sanitation. This was visible especially in the case when creating a payment order, and the review view was not visible. I tested this on the main branch as well, and the behaviour is the same.
- Added sanitation of amount when submitting feedback as well
- Updated tests to account for locale always being German.

Thank you!

PR #670, opened 2025-02-13
Finding Jira Ticket id in last three commits

obaidgini:
Added the logic to check last three commits if the jira ticket id is not found in last commit
PM-143


zladzeyka (approved):
thank you @obaidgini

PR #686, opened 2025-03-12
Ipc 608 porting the orders list from `MerchantSDK` example app to `HealthSDK` example app

danicretu:


health-sdk/example-app/src/main/java/net/gini/android/health/sdk/exampleapp/orders/OrdersActivity.kt
```diff
@@ -0,0 +1,226 @@
...
+            else -> super.onOptionsItemSelected(item)
+        }
+    }
+
+    companion object {
+        private val LOG = LoggerFactory.getLogger(OrdersActivity::class.java)
+    }
+}
+
+class OrdersAdapter(
```
obaidgini, line 187: Hi @danicretu instead of putting the adapters and extension function in this activity class, can we keep them in separate files? Or is there any specific reason behind it?  Just wanted to understand if we can separate them into separate files or not ?

health-sdk/example-app/src/main/java/net/gini/android/health/sdk/exampleapp/orders/OrdersViewModel.kt
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    fun startPaymentFlowWithoutDocument(paymentDetails: PaymentDetails) {
+        _startIntegratedPaymentFlow.tryEmit(paymentDetails)
+    }
+
+    fun getPaymentFragmentForPaymentDetails(paymentDetails: PaymentDetails, paymentFlowConfiguration: PaymentFlowConfiguration?): Result<PaymentFragment> {
+        try {
+            val paymentFragment = giniHealth.getPaymentFragmentWithoutDocument(paymentDetails, PaymentFlowConfiguration(shouldShowReviewBottomDialog = paymentFlowConfiguration?.shouldShowReviewBottomDialog ?: false, shouldHandleErrorsInternally = true))
+            return Result.success(paymentFragment)
```
obaidgini, line 62: @danicretu can you please check if it is formatted right? paymentFlowConfiguration?.shouldShowReviewBottomDialog

obaidgini (approved):
@danicretu Everything looks good to me 👌 , Tested on device as well, everything is looking good as per my understanding. Just a very small comment but it is not necessary. We can provide the build to QA and the points I have mentioned related to separating classes If you think it is not necessary then just comment on the PR.
Also we will ignore the sonar cloud warning for this PR as we know why it is happening and duplication is required.
Thank you.

MozhganPeivandianSharbaf (approved):

health-sdk/example-app/src/main/java/net/gini/android/health/sdk/exampleapp/orders/OrdersActivity.kt
danicretu: Hi @obaidgini , I've moved both the adapter and the extension function out of the `OrdersActivity`.

health-sdk/example-app/src/main/java/net/gini/android/health/sdk/exampleapp/orders/OrdersViewModel.kt
danicretu: @obaidgini , I added some clarifying parentheses so it's more clear what the parameter value should be.

PR #688, opened 2025-03-13
Ipc 606 expiring a payment request

danicretu:


health-sdk/example-app/src/main/java/net/gini/android/health/sdk/exampleapp/orders/OrderDetailsViewModel.kt
```diff
@@ -45,4 +58,51 @@ class OrderDetailsViewModel : ViewModel() {
...
+        )
+
+        if (paymentDetails.iban.isEmpty() || paymentDetails.amount.isEmpty() || paymentDetails.purpose.isEmpty() || paymentDetails.recipient.isEmpty()) {
+            _errorFlow.value = Error.PaymentDetailsIncomplete
+            return@launch
+        }
+        if (!isValidIban(paymentDetails.iban)) {
+            _errorFlow.value = Error.InvalidIban
+            return@launch
+        }
```
MozhganPeivandianSharbaf, line 77: [suggestion] Thanks for the great work on this task! I was thinking, maybe we could separate the validation logic into its own function? This way, createPaymentRequest would focus only on creating the request, and the responsibilities would be clearer. what do you think?

MozhganPeivandianSharbaf (approved):

obaidgini (approved):
Look good to me 👌 , Tested on the device as well, I can see the create payment request button working fine and expiry date is displayed as well, Thank you!

health-sdk/example-app/src/main/java/net/gini/android/health/sdk/exampleapp/orders/OrderDetailsViewModel.kt
danicretu: Thank you for the suggestion @MozhganPeivandianSharbaf ! I've extracted the validation logic to a separate method, please have a look.

PR #693, opened 2025-03-20
IPC-619 add stubs bulk delete payment requests

MozhganPeivandianSharbaf:


health-sdk/sdk/src/doc/source/flow.rst
```diff
@@ -85,6 +85,51 @@ into one final document:
     }
 
 Check which documents/invoices are payable
+---------------------------------
+
+``GiniHealthSDK`` provides a method to delete a payment request. You can do this by calling ``giniHealth.deletePaymentRequest(...)`` with a payment request ID.
```
danicretu, line 90: [Request] It seems that the sections on this page got a bit mixed up. For this header `Check which documents/invoices are payable` , the section is about deleting a payment request.

The `------------` marks a header section. Please update this one, and the ones below.

health-sdk/sdk/src/doc/source/flow.rst
```diff
@@ -85,6 +85,51 @@ into one final document:
...
+            is Resource.Error -> // Handle Error
+            is Resource.Cancelled -> //  Handle cancellation
+        }
+    }
+
+
+To delete a payment request, you must first create one to obtain a valid payment Request Id. Ensure that the request has not expired before attempting deletion.
+---------------------------------
+
+``GiniHealthSDK`` provides a  method to delete multiple payment request at once. You can do this by calling ``giniHealth.deletePaymentRequests(...)`` with a list of payment request IDs. The call will only succeed if all payment request were successfully deleted. If any payment request is invalid, unauthorized, or not found, the entire deletion request will fail, and no payment requests will be deleted. In the case of failures, an error or type ``DeletePaymentRequestErrorResponse`` will be provided, with more insight into why the deletion failed.
```
danicretu, line 111: [Request] Please add a header section for this method as well.

health-sdk/sdk/src/doc/source/flow.rst
```diff
@@ -85,6 +85,51 @@ into one final document:
...
+            is Resource.Success -> {
+                // `null` will be returned here
+            }
+            is Resource.Error -> // Handle `DeletePaymentRequestErrorResponse`
+            is Resource.Cancelled -> // Handle `DeletePaymentRequestErrorResponse`
+        }
+    }
+
+
+To check which payment requests are deletable, ensure that they are valid, authorized, and have not expired before attempting deletion.
```
danicretu, line 132: [Clarification] Is this necessary? This would imply that the client needs to fetch the payment request to check it before trying to delete it. I think that's why we have the different errors that the BE returns, so the client doesn't need to check beforehand. 
Also, not sure how one would check if a payment request is authorised 🤔

danicretu (changes requested):
Looks good, thank you! I added a couple of requests related to the documentation, it got a bit mixed up, but other than that, everything looks fine!

health-sdk/sdk/src/doc/source/flow.rst
MozhganPeivandianSharbaf: Thanks for catching that!

danicretu (approved):

PR #697, opened 2025-03-24
fix(health-sdk): fix showing error code

MozhganPeivandianSharbaf:
IPC-582

### DIFF
--- a/health-sdk/sdk/src/main/java/net/gini/android/health/sdk/GiniHealth.kt
+++ b/health-sdk/sdk/src/main/java/net/gini/android/health/sdk/GiniHealth.kt
@@ -117,7 +117,8 @@ class GiniHealth(
      */
     val openBankState: StateFlow<PaymentState> = _openBankState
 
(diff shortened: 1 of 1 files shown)


health-sdk/sdk/src/main/java/net/gini/android/health/sdk/GiniHealth.kt
```diff
@@ -206,7 +206,10 @@ class GiniHealth(
             )
         return when (response) {
             is Resource.Success -> null
-            is Resource.Error -> response.message
+            is Resource.Error -> response.message ?: response.responseStatusCode?.let {
```
danicretu, line 209: [Suggestion] I think this needs to check if `response.message.isEmpty()` 🤔
What do you think?

MozhganPeivandianSharbaf: yeah really good I was changing it thank you

danicretu (approved):
Thank you for the quick fix! 🔥

PR #700, opened 2025-03-26
Ipc 645 crash when clicking links on android 14 and 15

danicretu:


obaidgini (approved):
Looks good, Thank you!

PR #702, opened 2025-03-26
feat(health-sdk): expose getting payment

MozhganPeivandianSharbaf:
IPC-635


health-sdk/sdk/src/doc/source/flow.rst
```diff
@@ -169,6 +169,28 @@ Health API. You can then store the ``isPayable`` state in your own data model.
...
+        try {
+            // Retrieve payment details
+            val paymentDetails = giniHealth.getPayment(paymentId)
+        } catch (e: Exception) {
+            // Handle error
+        }
+    }
+
+
 Check if document has multiple invoices
```
danicretu, line 194: [Clarification] What do you think about refactoring the order of the sections a bit? 
The aim of the `flow` is to suggest a natural order in which to interact with our SDK.

Currently we have `Upload Document`, `Delete Payment Request`, `Delete multiple documents`, `Delete payment request`, `Check if document is payable`, `Get payment details`, `Check if document has multiple invoices`, `Create the payment fragment`.

My 2 thoughts here:

- Why do we have `Delete Payment Request` twice here? 🤔 Perhaps something happened during some merge
- In order to preserve the clarity of the flow, I would suggest moving the documentation for all the `delete` methods to be last in page, before the `PaymentFlowConfiguration` class details. I would also move the `Get Payment Details` after `Create Payment Fragment`, since checking whether a document has multiple invoices and whether it is payable should happen at the same time (it doesn't make much sense to instruct the developer to check if a document is payable, then to get the payment details and then to check if the document has multiple invoices).

The main idea I am trying to convey here is that we should aim to keep the order in this documentation as close to the order of the steps that the client needs to take to interact with our SDK.

What are your thoughts on this? I am happy to discuss more if needed.

danicretu (changes requested):
Great work, thanks! 

I added just a small request to start the section with an uppercase, and a discussion about the order of the sections in the documentation.

Please let me know if anything is unclear, I am happy to discuss.

danicretu (approved):
Thank you for making the suggested changes, everything looks good!

PR #705, opened 2025-04-01
PP-1105 : Updated the targetSdk level to 35

obaidgini:
PP-1105


abolfazlimahdi (approved):
Looks good. Thank you!

PR #708, opened 2025-04-08
fix(health-sdk): Fix app crash when expirationDate is not null, but e…

danicretu:
…mpty

IPC-644


obaidgini (approved):
Look good, Thank you.

MozhganPeivandianSharbaf (approved):
[suggestion] Thank you Dani it looks good to me just one small thing, Nullability and empty string checks for isInTheFuture were implemented in the bulk deletion payment request branch, you can delete it or we can fully address it once the branch is merged.

danicretu:
Thank you @MozhganPeivandianSharbaf for pointing it out. I merged that branch in this one, because there was still someplace left where there was an `isEmpty` check missing.

PR #715, opened 2025-04-15
feature(bank-sdk): Skonto Invoice Preview. Error handling

ndugini:
PP-740


abolfazlimahdi (changes requested):
[Request] I see that you have added the functionality for the TD but the first request is not met yet. Please let me know if there are problems with reusing the icon from Capture SDK. Thank you!

1. Please use gc_error_server_icon drawable from capture-sdk instead of adding a new icon

PR #724, opened 2025-04-28
feat(health-sdk): Change input IBAN input field from lowercase to upp…

MozhganPeivandianSharbaf:
…ercase automatically

IPC-658

### DIFF
--- a/internal-payment-sdk/sdk/src/main/java/net/gini/android/internal/payment/review/reviewComponent/ReviewView.kt
+++ b/internal-payment-sdk/sdk/src/main/java/net/gini/android/internal/payment/review/reviewComponent/ReviewView.kt
@@ -1,8 +1,10 @@
 package net.gini.android.internal.payment.review.reviewComponent
 
 import android.content.Context
+import android.text.Editable
 import android.text.SpannableStringBuilder
 import android.text.Spanned
+import android.text.TextWatcher
 import android.text.style.ImageSpan
 import android.util.AttributeSet
 import android.view.View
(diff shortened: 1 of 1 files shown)


obaidgini (changes requested):
Hi @MozhganPeivandianSharbaf I have a very small comment, what you are doing manually can be tackled by the system automatically if you will put android:inputType="text|textCapCharacters" in gps_review.xml in the text input with id android:id="@+id/iban"
Adventages :
1- You don't need to handle the uppercase manually.
2- You don't have to add/remove listeners 
3- You don't need to worry about selections
4- The keyboard on the mobile will show the user that letters will be typed in caps

MozhganPeivandianSharbaf:
> Hi @MozhganPeivandianSharbaf I have a very small comment, what you are doing manually can be tackled by the system automatically if you will put android:inputType="text|textCapCharacters" in gps_review.xml in the text input with id android:id="@+id/iban" Adventages : 1- You don't need to handle the uppercase manually. 2- You don't have to add/remove listeners 3- You don't need to worry about selections 4- The keyboard on the mobile will show the user that letters will be typed in caps

Hi @obaidgini  thank you for you suggestion,I initially considered inputType="text|textCapCharacters", but I noticed that it only suggests uppercase input via the keyboard — it doesn't enforce it. For example, if a user:

Pastes lowercase content

Uses a physical keyboard

Or changes the keyboard input mode manually

…then the field may still contain lowercase characters.

Since the requirement was to guarantee that IBANs are uppercased regardless of input source, I used a TextWatcher to ensure this transformation is applied uniformly. It’s a bit more code, but it's more robust for this case.

Let me know if you’d still recommend another approach

obaidgini:
No, you are right. @MozhganPeivandianSharbaf 
- Uses a physical keyboard
- Or changes the keyboard input mode manually

These cases are very important, i will approve it now, Well done for thinking about every use case :)

MozhganPeivandianSharbaf:
Thanks a lot @obaidgini 😊 ! Glad we aligned on it. 🙌

PR #752, opened 2025-05-26
feature(capture-sdk): Invoice education. Implement Edge Cases

ndugini:
PP-1235


abolfazlimahdi (changes requested):
[Request] App crashes when no result screen should be shown.

PR #756, opened 2025-05-26
fix(bank-sdk): enhance accessibility by adding focusable attributes 

MozhganPeivandianSharbaf:
PP-1205


abolfazlimahdi (approved):
Looks good. Thank you!

PR #777, opened 2025-06-12
feat(bank-sdk): fixing the scroll issues in RA

obaidgini:
PP-1432


abolfazlimahdi (approved):
Please make sure you don't need to fix the same issue on different devices. Other than that, it looks good. Thanks

PR #783, opened 2025-06-24
Ipc 666 200 percent text support

obaidgini:


health-sdk/example-app/src/main/res/xml/locale_configs.xml
MozhganPeivandianSharbaf, line 1: Looks good! Just one thing: since we're supporting three languages (English, German formal and informal), and using different resource folders (values, values-de, and values-en), maybe we should explicitly define both German variants in locale_config.xml using de-DE and de-DE-u-formal. This helps Android distinguish between the two, especially from API 33+.

MozhganPeivandianSharbaf, line 1: Looks good! Just one suggestion: since we're supporting three languages (English, German formal and informal), and using different resource folders (values, values-de, and values-en), maybe we should explicitly define both German variants in locale_config.xml using de-DE and de-DE-u-formal. This helps Android distinguish between the two, especially from API 33+.

MozhganPeivandianSharbaf (changes requested):

health-sdk/example-app/src/main/res/xml/locale_configs.xml
MozhganPeivandianSharbaf: @obaidgini  Got it, that makes sense! Thanks for clarifying that the formal/informal handling is managed on the server side — so even if Android offers multiple German options, the config API will always determine the actual variant used. Sounds like for now, keeping only de-DE and en in locale_config.xml is enough. 👍

MozhganPeivandianSharbaf (approved):
Looks good, Thank you!

PR #787, opened 2025-07-01
fix(internal-payment-sdk): adding new strings and text for bank selec…

obaidgini:
…tion

IPC-737


abolfazlimahdi (approved):
Looks good. Thanks

PR #790, opened 2025-07-07
IPC 721 keyboard issues on xiaomi device with android 14

MozhganPeivandianSharbaf:
…ures

IPC-721


health-sdk/sdk/src/main/java/net/gini/android/health/sdk/integratedFlow/PaymentFragment.kt
```diff
@@ -346,23 +346,6 @@ class PaymentFragment private constructor(
...
-                            if (viewModel.documentId != null) {
-                                showReviewFragment()
-                                return
-                            }
-                            else if (viewModel.paymentFlowConfiguration?.shouldShowReviewBottomDialog == true) {
-                                showReviewBottomDialog()
-                                return
-                            }
-                    }
-
```
obaidgini, line 365: Is this code removed by mistake? because it have the logic which checks that review bottom sheet is turned on in the config or not? can you please elaborate this?

MozhganPeivandianSharbaf: yes this logic added By Dani and we are not sure why it added in physical keyboard issues and created some bug for review bottom sheet and its condition was not correct. I asked her she has some doubt I removed it and the issue fixed and I tested flow everything was fine.
but I will check with Saranya during QA testing to make sure.

obaidgini: Ok, Thank you, Please take care of this when you are testing with Saranya. Thanks

PR #812, opened 2025-07-29
IPC 766 fix lifecycle structure for all screens and bottom sheet when app was killed by the system

MozhganPeivandianSharbaf:
…tom sheet WHEN app was killed by the system

IPC-766


obaidgini (approved):
Looks good @MozhganPeivandianSharbaf 
Thank you for adjusting all the suggestions from mahdi and my side 👌

PR #836, opened 2025-09-08
PP-1480: Integration of Capture SDK

obaidgini:
### Description

Here is the [Jira Ticket Link](https://ginis.atlassian.net/browse/PP-1480)

### Notes for reviewer

Fixes crash in the example app, when GiniCapture SDK is started.

**Reason detected:** 
Android system was not able to create GiniCaptureFragment because this class have parameterized constructor, So fragment factory is provided in the example app, which Android system will use to construct the fragment in case of orientation change.

Also, changed the `lateinit giniCaptureFragmentListener` property of `GiniCaptureFragment` to nullable, so it can be marked as null when the `onDestroy` is called, and it can be collected by GC.
Because we are setting the listener again in the event of orientation change through `ClientCaptureSDKFragmentFactory`, so we will keep getting the updates.

Thank you.

### DIFF
--- a/bank-sdk/example-app/src/main/java/net/gini/android/bank/sdk/exampleapp/ui/CaptureSdkStandAloneActivity.kt
+++ b/bank-sdk/example-app/src/main/java/net/gini/android/bank/sdk/exampleapp/ui/CaptureSdkStandAloneActivity.kt
@@ -4,39 +4,83 @@ import android.content.Context
 import android.content.Intent
 import android.os.Bundle
 import androidx.appcompat.app.AppCompatActivity
+import androidx.fragment.app.Fragment
+import androidx.fragment.app.FragmentFactory
 import dagger.hilt.android.AndroidEntryPoint
-import net.gini.android.bank.sdk.di.BankSdkIsolatedKoinContext
 import net.gini.android.bank.sdk.exampleapp.R
 import net.gini.android.bank.sdk.exampleapp.core.ExampleUtil.isIntentActionViewOrSend
 import net.gini.android.bank.sdk.exampleapp.ui.util.CaptureResultListener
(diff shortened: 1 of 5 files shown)


bank-sdk/example-app/src/main/java/net/gini/android/bank/sdk/exampleapp/ui/CaptureSdkStandAloneActivity.kt
```diff
@@ -4,39 +4,83 @@ import android.content.Context
...
-        // same view is required for both Bank SDK and Capture SDK, no need to create a separate xml
-        setContentView(R.layout.activity_capture_flow_host)
+        setContentView(R.layout.activity_stand_alone_capture_flow)
         if (savedInstanceState == null) {
             if (intent != null && isIntentActionViewOrSend(intent)) {
                 ClientCaptureSDKFragment().startCaptureSDKForIntent(
                     this,
                     intent,
-                    CaptureResultListener(this)
+                    listener!!
```
abolfazlimahdi, line 33: [Request] I know this is inside the example app but using double bang is not recommended. Can you double check if you can have any work arounds for this?

abolfazlimahdi (approved):
I have added a comment. Please take a look and if applicable, try to fix it. Thank you!

bank-sdk/example-app/src/main/java/net/gini/android/bank/sdk/exampleapp/ui/CaptureSdkStandAloneActivity.kt
abolfazlimahdi: Got it. Thanks for the clarification.

PR #840, opened 2025-09-18
PP 1474 Invoice marked as paid - Parse the new payment sta…

MozhganPeivandianSharbaf:
…te object

PP-1474

### DIFF
--- a/bank-api-library/library/src/main/java/net/gini/android/bank/api/response/ConfigurationResponse.kt
+++ b/bank-api-library/library/src/main/java/net/gini/android/bank/api/response/ConfigurationResponse.kt
@@ -15,6 +15,7 @@ data class ConfigurationResponse(
     @Json(name = "qrCodeEducationEnabled") val qrCodeEducationEnabled: Boolean?,
     @Json(name = "instantPaymentEnabled") val instantPaymentEnabled: Boolean?,
     @Json(name = "eInvoiceEnabled") val eInvoiceEnabled: Boolean?,
+    @Json(name = "paymentHintsEnabled") val paymentHintsEnabled: Boolean?,
 )
 
 internal fun ConfigurationResponse.toConfiguration() = Configuration(
@@ -26,6 +27,7 @@ internal fun ConfigurationResponse.toConfiguration() = Configuration(
     transactionDocsEnabled = transactionDocsEnabled ?: false,
     qrCodeEducationEnabled = qrCodeEducationEnabled ?: false,
     instantPaymentEnabled = instantPaymentEnabled ?: false,
-    isEInvoiceEnabled = eInvoiceEnabled ?: false
+    isEInvoiceEnabled = eInvoiceEnabled ?: false,
+    paymentHintsEnabled =  paymentHintsEnabled ?:false
 )
 
--- a/bank-sdk/example-app/src/main/res/values/strings.xml
+++ b/bank-sdk/example-app/src/main/res/values/strings.xml
@@ -59,6 +59,7 @@
     <string name="relaunch_to_use_default_configuration">Please relaunch the app to use the default GiniConfiguration values.</string>
     <string name="close_screen_for_configurations_to_take_place">The configuration changes will take effect after closing this screen.</string>
     <string name="default_sdk_configuration_switch_label">SDK default configurations</string>
+    <string name="paid_invoice_switch_label">Paid invoice warning</string>
     <string name="open_with_switch_label">Open with</string>
     <string name="qr_code_scanning_switch_label">QR code scanning</string>
     <string name="only_qr_code_scanning_switch_label">QR code scanning only</string>
--- a/capture-sdk/sdk/src/main/java/net/gini/android/capture/analysis/AnalysisScreenPresenter.java
+++ b/capture-sdk/sdk/src/main/java/net/gini/android/capture/analysis/AnalysisScreenPresenter.java
@@ -10,6 +10,8 @@ import net.gini.android.capture.AsyncCallback;
 import net.gini.android.capture.Document;
 import net.gini.android.capture.GiniCapture;
 import net.gini.android.capture.GiniCaptureError;
+import net.gini.android.capture.analysis.warning.WarningPaymentState;
+import net.gini.android.capture.analysis.warning.WarningType;
 import net.gini.android.capture.document.DocumentFactory;
 import net.gini.android.capture.document.GiniCaptureDocument;
 import net.gini.android.capture.document.GiniCaptureDocumentError;
@@ -25,6 +27,7 @@ import net.gini.android.capture.internal.storage.ImageDiskStore;
 import net.gini.android.capture.internal.util.FileImportHelper;
 import net.gini.android.capture.logging.ErrorLog;
 import net.gini.android.capture.logging.ErrorLogger;
+import net.gini.android.capture.network.model.GiniCaptureSpecificExtraction;
 import net.gini.android.capture.tracking.AnalysisScreenEvent;
 import net.gini.android.capture.tracking.AnalysisScreenEvent.ERROR_DETAILS_MAP_KEY;
 
@@ -55,6 +58,9 @@ class AnalysisScreenPresenter extends AnalysisScreenContract.Presenter {
 
     @VisibleForTesting
     static final String PARCELABLE_MEMORY_CACHE_TAG = "ANALYSIS_FRAGMENT";
+
+    private static final String EXTRACTION_PAYMENT_STATE = "paymentState";
+
     private static final Logger LOG = LoggerFactory.getLogger(AnalysisScreenPresenter.class);
 
     @VisibleForTesting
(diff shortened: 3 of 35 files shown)


bank-api-library/library/src/main/java/net/gini/android/bank/api/response/ConfigurationResponse.kt
```diff
@@ -15,6 +15,7 @@ data class ConfigurationResponse(
     @Json(name = "qrCodeEducationEnabled") val qrCodeEducationEnabled: Boolean?,
     @Json(name = "instantPaymentEnabled") val instantPaymentEnabled: Boolean?,
     @Json(name = "eInvoiceEnabled") val eInvoiceEnabled: Boolean?,
+    @Json(name = "paymentHintsEnabled") val paymentHintsEnabled: Boolean?,
 )
```
obaidgini, line 19: [Comment I believe if we have same variable name we don't need to have @Json(name = "paymentHintsEnabled"), or is this different for this class?

bank-sdk/example-app/src/main/res/values/strings.xml
```diff
@@ -110,6 +111,7 @@
     <string name="custom_primary_compose_button_switch_label">Custom primary button in Skonto</string>
     <string name="return_assistant_feature_switch_label">Return assistant feature</string>
     <string name="return_assistant_feature_switch_description">Present a digital representation of the invoice</string>
+    <string name="payment_hint_switch_description">Toggle to show a warning when an invoice is marked as paid or similar states</string>
     <string name="skonto_feature_switch_label">Skonto feature</string>
```
obaidgini, line 115: [Comment] : Do we need to take care of German translation or it is not needed?

capture-sdk/sdk/src/main/java/net/gini/android/capture/analysis/warning/WarningBottomSheet.kt
```diff
@@ -0,0 +1,172 @@
...
+
+    private var titleText: CharSequence? = null
+    private var descText: CharSequence? = null
+
+    var listener: Listener? = null
+
+    @Suppress("DEPRECATION")
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        val type = arguments?.getSerializable(ARG_TYPE) as? WarningType
```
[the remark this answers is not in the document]
obaidgini: +1 for this comment

capture-sdk/sdk/src/main/java/net/gini/android/capture/di/CaptureSdkIsolatedKoinContext.kt
```diff
@@ -11,7 +11,8 @@ object CaptureSdkIsolatedKoinContext {
             educationModule,
             qrEducationModule,
             invoiceEducationModule,
-            EInvoiceModule
+            EInvoiceModule,
+            PaymentHintsModule
         )
```
obaidgini, line 16: [Request]: The naming convention is looking inconsistent here, for example invoiceEducationModule and PaymentHintsModule
for invoiceEducationModule camelCase is used
and for EInvoiceModule and PaymentHintsModule pascal case is used, is there any specific reason or this is by mistake?

capture-sdk/sdk/src/main/java/net/gini/android/capture/di/PaymentHintsModule.kt
```diff
@@ -0,0 +1,14 @@
+package net.gini.android.capture.di
+
+import net.gini.android.capture.paymentHints.GetPaymentHintsEnabledUseCase
+import org.koin.dsl.module
+
+internal val PaymentHintsModule = module {
+
```
obaidgini, line 7: This is pascal case naming, and in other modules we are using camelCase

capture-sdk/sdk/src/main/java/net/gini/android/capture/internal/ui/FragmentImplCallback.java
```diff
@@ -47,6 +47,8 @@ void showAlertDialog(@NonNull final String message,
     @NonNull
     FragmentManager getChildFragmentManager();
 
+    default void showWarning(@NonNull WarningType warningType, @NonNull Runnable onProceed) {}
```
obaidgini, line 50: [Request]: Is there any specific reason we are using default keyword here? because we are not providing any implementation in the body of this method?

capture-sdk/sdk/src/main/res/values/dimens.xml
```diff
@@ -48,5 +48,4 @@
     <dimen name="gc_unknown_banner_width">330dp</dimen>
     <dimen name="gc_unelevated_button_height">55dp</dimen>
     <dimen name="gc_snackbar_text_height">60dp</dimen>
-
 </resources>
```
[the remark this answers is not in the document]
obaidgini: +1

capture-sdk/sdk/src/main/java/net/gini/android/capture/analysis/AnalysisScreenPresenter.java
```diff
@@ -311,11 +318,19 @@ public Void apply(final AnalysisInteractor.ResultHolder resultHolder,
...
+                                } else if (shouldShowPaidInvoiceWarning(resultHolder)) {
+                                    getView().showPaidWarningThen(
+                                            WarningType.DOCUMENT_MARKED_AS_PAID,
+                                            () -> proceedWithExtractions(resultHolder)
+                                    );
                                 } else {
                                     proceedWithExtractions(resultHolder);
                                 }
+                                break;
+                            }
```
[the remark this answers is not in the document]
MozhganPeivandianSharbaf: Good point @obaidgini I will Rollback this change to work on it in separate, related ticket to test it carefully.

capture-sdk/sdk/src/main/java/net/gini/android/capture/di/CaptureSdkIsolatedKoinContext.kt
MozhganPeivandianSharbaf: I used camelCase based on EInvoiceModule I will change it

bank-api-library/library/src/main/java/net/gini/android/bank/api/response/ConfigurationResponse.kt
MozhganPeivandianSharbaf: for sure, as you mentioned there is no need to have @Json(name =) for none of variable because the have same name I use it to have consistency with other fields

capture-sdk/sdk/src/main/java/net/gini/android/capture/analysis/warning/WarningPaymentState.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+ public enum WarningPaymentState {
+    PAID, TO_BE_PAID, BOOKED, UNKNOWN;
+
+    public static WarningPaymentState from(@Nullable String raw) {
+        if (raw == null) return UNKNOWN;
+        switch (raw.trim()) {
+            case "Paid":      return PAID;
+            case "ToBePaid":  return TO_BE_PAID; // in future versions this might be "ToBePaid"
+            case "booked":    return BOOKED; // in future versions this might be "Booked"
```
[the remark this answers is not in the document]
MozhganPeivandianSharbaf: this class is for Warning ,for now we just implemented warning of Paid state and other states will implemented in next user stories, I will change comments to make it more clear

capture-sdk/sdk/src/main/java/net/gini/android/capture/analysis/warning/WarningBottomSheet.kt
```diff
@@ -0,0 +1,172 @@
...
+
+    private lateinit var binding: GcWarningBottomSheetBinding
+
+    private var titleText: CharSequence? = null
+    private var descText: CharSequence? = null
+
+    var listener: Listener? = null
+
+    @Suppress("DEPRECATION")
+    override fun onCreate(savedInstanceState: Bundle?) {
```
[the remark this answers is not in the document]
MozhganPeivandianSharbaf: Good catch!
The original code was using getSerializable(ARG_TYPE), which is deprecated on API 33+. That’s why I had temporarily suppressed the warning.

I’ve updated the code so that on Android 13+ (API 33 and above) we now use the new overload getSerializable(ARG_TYPE, WarningType::class.java). For older devices we still need the old method, so I left a fallback with @Suppress("DEPRECATION") just for API < 33.

Since WarningType is an enum (and enums are already Serializable), this is the simplest and correct approach. We don’t need to make it Parcelable just for this.

So yes, we are now using the new method where available, and the deprecation suppression is only kept for backward compatibility with our minSdk (21).

obaidgini (changes requested):
[Request]: Can you please use the camelCase for EInvoiceModule as well in EInvoiceModule.kt ? Thanks.

PR #841, opened 2025-09-24
PP 1761 blocker:  AnalysisScreenPresenter issues

MozhganPeivandianSharbaf:
…ch case

PP-1761


obaidgini (approved):
Looks good to me @MozhganPeivandianSharbaf, I don't think there is any issue in this small change, as the blocks where you added the break, after that there is no logic other then breaking the flow, so over all I believe there will be no difference in the logic. 
Thank you.

PR #843, opened 2025-09-24
PP 1763 DocumentAnalyzer screen issues

MozhganPeivandianSharbaf:
PP-1763


bank-sdk/example-app/src/main/java/net/gini/android/bank/sdk/exampleapp/core/DocumentAnalyzer.kt
```diff
@@ -19,6 +19,9 @@ class DocumentAnalyzer @JvmOverloads internal constructor(
     private val documentManager: DocumentManager<*, *>,
     coroutineContext: CoroutineContext = Dispatchers.Main
 ) {
+    companion object {
+        private const val LOG_TAG = "gini-api"
+    }
```
obaidgini, line 24: @MozhganPeivandianSharbaf Can you please move this companion object to the end of this class? because usually we are doing this, and it will keep the code style consistent. Thank you

PR #846, opened 2025-09-26
IPC 746 implement MainScreen composable

MozhganPeivandianSharbaf:
I re-implemented the `MainActivity` of the Health SDK using Jetpack Compose. This was part of applying the new concepts and techniques I learned during my Compose course.

Instead of relying on XML layouts and ViewBinding, I moved the entire UI into a `MainScreen()` composable. This allowed me to:

- Replace `ConstraintLayout` and `ViewPager2` with Compose layouts and the new HorizontalPager API.
- Use `rememberLauncherForActivityResult` to handle camera capture and document import in a declarative way.
- Collect `Flow` state directly in Compose using `collectAsState`, so the UI automatically updates when the ViewModel state changes.
- Add a custom dots indicator to replace the `TabLayoutMediator`.
- Simplify button enable/disable logic by binding directly to state (Upload becomes active as soon as pages are present).

Through this exercise, I could demonstrate what I learned in the course: **writing modern, declarative UI, managing lifecycle safely with Compose, and reducing boilerplate compared to XML + imperative View setup.**

It was also a chance to share with the team how our existing SDK examples can be updated to Compose-first architecture, which makes the codebase cleaner, more maintainable, and aligned with current Android best practices.

### DIFF
--- a/health-sdk/example-app/src/main/java/net/gini/android/health/sdk/exampleapp/MainActivity.kt
+++ b/health-sdk/example-app/src/main/java/net/gini/android/health/sdk/exampleapp/MainActivity.kt
@@ -1,28 +1,19 @@
 package net.gini.android.health.sdk.exampleapp
 
 import android.content.Intent
-import android.net.Uri
 import android.os.Bundle
-import android.widget.Toast
 import androidx.activity.compose.setContent
-import androidx.activity.result.contract.ActivityResultContracts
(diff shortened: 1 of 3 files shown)


abolfazlimahdi (changes requested):
[Request] Added two comments to the PR. Overall, it looks good :)

PR #883, opened 2026-02-17
Pp 2078 provide color keys per element

abolfazlimahdi:
Atruvia supports 25–30 color themes for their banking customers. They currently manually override colors for multiple UI components. They requested more (ideally all) components be color-customizable to speed up publishing the photo payment module.

This is my proposal for providing color keys per element. 

Transition to Element-Specific Color Keys (XML)
In commit c0e5abe12, the project moved away from using generic color primitives directly in layouts. Previously, UI elements like the camera title in gc_fragment_camera.xml were hardcoded to use general colors like @color/gc_light_01.
The update introduced element_colors.xml, which defines semantic color keys such as gc_camera_title_color. This allows developers to theme specific UI elements without affecting others that might share the same base color. For example, changing the camera title color no longer requires changing the global gc_light_01 value, providing much better flexibility for white-labeling and customization.

Integration with Jetpack Compose
In commit 76943d8e9, this semantic approach was extended to the Compose-based elements. A new class, GiniColorElementsPrimitives, was introduced to act as a bridge. It reads these new semantic XML color keys and maps them into the Compose GiniColorScheme.
Key technical changes include:
• GiniColorElementsPrimitives: A new data class that holds element-specific colors (e.g., skontoSectionTitleTextColor) and initializes them from Android resources.
• Theme Expansion: The GiniColorScheme and GiniTheme were updated to include a SkontoSection color group. This ensures that even the most specific UI components (like the Skonto section titles) are now part of the unified theme state.
• Dynamic Updating: The giniLightColorScheme and giniDarkColorScheme functions now accept these element primitives, ensuring that both light and dark modes respect the new specific color overrides.

How the clients can use:

For clients who are used to the old color mapping: no changes will happen. They can still override our color palette and customize our SDK.

For clients who need more detailed and customizable SDK, they need to override the elements that they need. This still uses our color palette and if clients don’t override some colors, we’ll use the colors from our color palette.

### DIFF
--- a/bank-sdk/example-app/src/main/res/values/colors.xml
+++ b/bank-sdk/example-app/src/main/res/values/colors.xml
@@ -2,6 +2,7 @@
 <resources>
 
     <color name="gc_camera_title_color">#006ECF</color>
+    <color name="skonto_section_title_text_color">#006ECF</color>
 
     <color name="gc_color_01">@color/gc_light_01</color>
     <color name="gc_color_02">@color/gc_light_02</color>
--- a/bank-sdk/example-app/src/main/res/values/colors.xml
+++ b/bank-sdk/example-app/src/main/res/values/colors.xml
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
+
+    <color name="gc_camera_title_color">#006ECF</color>
+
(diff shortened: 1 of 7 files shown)


abolfazlimahdi:
Not needed anymore

PR #914, opened 2026-05-04
Pp 2508 bs automation

abolfazlimahdi:


bank-sdk/example-app/src/androidTest/java/net/gini/android/bank/sdk/exampleapp/ui/screens/ExtractionScreen.kt
```diff
@@ -5,23 +5,40 @@ import androidx.test.espresso.action.ViewActions.click
...
+    // The extraction screen only appears after the Gini API returns results, which can
+    // take noticeably longer on remote/BrowserStack devices than the fixed IdlingResource
+    // sleep. Wait for the transfer-summary button to actually exist before interacting,
+    // so slow network responses don't cause a NoMatchingViewException.
+    private fun waitForExtractionScreen() {
+        val device = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
+        device.findObject(
+            UiSelector().resourceId("net.gini.android.bank.sdk.exampleapp:id/transfer_summary")
+        ).waitForExists(EXTRACTION_TIMEOUT)
+    }
```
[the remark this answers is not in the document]
MozhganPeivandianSharbaf: Good catch — fixed. I now resolve the package at runtime via `InstrumentationRegistry.getInstrumentation().targetContext.packageName` instead of hard-coding it, in both `ExtractionScreen.waitForExtractionScreen()` and `DigitalInvoiceEditButton.checkEditButtonTitleIsDisplayed()`, so the selector stays correct across flavors.

For context: the UI suite currently only builds/runs on the `devExampleApp` flavor (whose applicationId matches the old literal), and this hard-coded pattern already exists in several pre-existing page objects — but making the new code flavor-safe is the right call. Thanks!

bank-sdk/example-app/src/androidTest/scripts/bs_build_and_upload.sh
```diff
@@ -50,104 +70,140 @@ if [[ "$BS_USER" == "<your_browserstack_user_name>" ]] || [[ "$BS_KEY" == "<your
...
+if [ -n "$ARTIFACT_URLS_FILE" ]; then
+  {
+    echo "APP_URL=$APP_URL"
+    echo "TEST_URL=$TEST_URL"
+    echo "IMAGE_URL=$IMAGE_URL"
+    echo "PDF_URL=$PDF_URL"
+    echo "SAMPLE_PDF_URL=$SAMPLE_PDF_URL"
+  } > "$ARTIFACT_URLS_FILE"
+  echo "Wrote artifact URLs to $ARTIFACT_URLS_FILE"
 fi
```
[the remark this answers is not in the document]
MozhganPeivandianSharbaf: Fixed. `bs_run_all_groups.sh` no longer `source`s the artifact-URLs file — it now parses each value with `grep`/`cut` and exports it explicitly, so the file's contents are never executed as shell code. The values are `bs://…` URLs returned by BrowserStack's API (no shell-special characters), but avoiding `source` removes the risk entirely. Thanks!

bank-sdk/example-app/src/androidTest/scripts/bs_build_and_upload.sh
```diff
@@ -163,6 +219,10 @@ BUILD_RESPONSE=$(curl -s -u "$BS_USER:$BS_KEY" \
     \"devices\": [\"$DEVICE_1\", \"$DEVICE_2\"],
     \"app\": \"$APP_URL\",
     \"testSuite\": \"$TEST_URL\",
+    \"singleRunnerInvocation\": \"true\",
+    \"useOrchestrator\": \"true\",
+    \"clearPackageData\": \"true\",
+    \"disableAnimations\": \"true\",
```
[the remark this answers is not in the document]
MozhganPeivandianSharbaf: Keeping these as strings intentionally. BrowserStack's App Automate Espresso v2 build API documents these flags as strings (e.g. `"true"`), not JSON booleans — and they're confirmed working in our runs: `clearPackageData: "true"` gave us the per-test isolation that fixed onboarding-state leakage, and `disableAnimations: "true"` resolved an animation-related `PerformException`. Neither fix would have taken effect if the string values were being ignored. Encoding them as booleans would risk BrowserStack not applying them, so no change here.

bank-sdk/example-app/src/androidTest/java/net/gini/android/bank/sdk/exampleapp/ui/resources/RetryRule.kt
```diff
@@ -0,0 +1,60 @@
...
+ * The Bank SDK UI tests exercise real end-to-end flows that call the Gini API over the
+ * network. On remote/BrowserStack devices these occasionally lose a timing race and fail
+ * transiently. Retrying gives such a test another chance, while a genuinely broken test
+ * (which fails on every attempt) still fails.
+ *
+ * Apply it as the OUTERMOST rule (lowest `order`, since in JUnit a lower order wraps
+ * further out) so each attempt re-runs the whole test — including the ActivityScenarioRule,
+ * which relaunches a fresh activity:
+ *
+ *     @get:Rule(order = -1)
```
abolfazlimahdi: **Blocking:** `order = -1` is `Rule.DEFAULT_ORDER`, so this is identical to not specifying an order at all — the retry rule *ties* with `activityScenarioRule` instead of wrapping it.

For tied `@get:Rule` (method) rules, JUnit falls back to `MethodSorter.DEFAULT`, which orders by method-name hash — effectively arbitrary. If the tie-break puts `RetryRule` *inside* `ActivityScenarioRule`, a retry re-runs the test body against the already-used activity instead of relaunching it, which defeats the rule's documented purpose.

Suggestion: use an order strictly below the default, e.g. `@get:Rule(order = -2)` (or `Int.MIN_VALUE`), update this KDoc example, and adjust the usages in the test classes accordingly. `OpenWithTest` uses a plain `@get:Rule` — harmless there since it has no activity rule, but worth aligning for consistency.

bank-sdk/example-app/src/androidTest/java/net/gini/android/bank/sdk/exampleapp/ui/testcases/ErrorScreenTests.kt
```diff
@@ -84,6 +93,13 @@ class ErrorScreenTests {
     @Test
     fun test2_verifyNetworkErrorScreen() {
         ErrorScreen().disconnectTheInternetConnection()
+        // On devices where the network can't actually be turned off (e.g. BrowserStack cloud
+        // devices, where `svc wifi/data disable` is blocked), skip: the offline scenario
+        // cannot be exercised there. Runs normally on local devices/emulators.
+        Assume.assumeFalse(
+            "Skipping: network could not be disabled on this device (e.g. BrowserStack)",
+            errorScreen.isInternetAvailable()
+        )
```
abolfazlimahdi: **Blocking:** only `test2` got this guard, but `test3_navigateToMainScreenByClickingEnterManuallyButton` and `test4_navigateToCameraScreenByClickingBackToCameraButton` rely on the same `svc wifi/data disable` working. On BrowserStack they'll run with the network up — either failing (and now burning 3 attempts each via `RetryRule` first), or passing for the wrong reason (e.g. "Enter manually" also exists on the no-result screen).

Please apply the same `Assume.assumeFalse(errorScreen.isInternetAvailable())` guard to both, or document why they're exempt.

Separately (pre-existing, non-blocking): `reconnectTheInternetConnection()` runs at the end of the test body, so any assertion failure leaves the device offline for subsequent tests — moving it to an `@After` would be more robust.

bank-sdk/example-app/src/androidTest/java/net/gini/android/bank/sdk/exampleapp/ui/screens/ErrorScreen.kt
```diff
@@ -10,6 +13,17 @@ import androidx.test.uiautomator.UiSelector
...
 
+    // Whether the device currently has an internet-capable active network. Used to detect
+    // environments (e.g. BrowserStack cloud devices) where disconnectTheInternetConnection()
+    // cannot actually turn the network off, so the offline test can be skipped there.
+    fun isInternetAvailable(): Boolean {
+        val context = InstrumentationRegistry.getInstrumentation().targetContext
+        val cm = context.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
+        val network = cm.activeNetwork ?: return false
+        val caps = cm.getNetworkCapabilities(network) ?: return false
+        return caps.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
```
abolfazlimahdi, line 38: Minor timing race: this is evaluated ~2s after `svc wifi disable`, and on a slow local device ConnectivityManager can still briefly report the dying network as available — which would wrongly *skip* the test on a device where the disconnect actually works.

A short poll loop (e.g. up to ~5s waiting for connectivity to drop) before deciding to skip would make this reliable.

bank-sdk/example-app/src/androidTest/java/net/gini/android/bank/sdk/exampleapp/ui/screens/DigitalInvoiceScreen.kt
```diff
@@ -38,7 +38,7 @@ class DigitalInvoiceScreen {
                 .text("Digital invoice")
                 .resourceId("net.gini.android.bank.sdk.exampleapp:id/onboarding_text_1")
         )
-        return onboardingScreenText.waitForExists(8000)
+        return onboardingScreenText.waitForExists(ONBOARDING_TIMEOUT)
```
abolfazlimahdi, line 42: The selectors in this class still hard-code `net.gini.android.bank.sdk.exampleapp:id/...`, while this PR fixes exactly that in `ExtractionScreen` and `DigitalInvoiceEditButton` by resolving `targetContext.packageName` at runtime (the applicationId varies by flavor).

The same flavor problem applies here — worth fixing in this PR too, ideally via a small shared helper so the resolution logic lives in one place.

bank-sdk/example-app/src/androidTest/java/net/gini/android/bank/sdk/exampleapp/ui/testcases/OpenWithTest.kt
```diff
@@ -15,26 +16,37 @@ import androidx.test.ext.junit.runners.AndroidJUnit4
...
+
     private var idlingResourceForOpenWith: IdlingResource? = null
 
     @Before
     fun setup() {
+        // Multi-image open-with uploads several files over the network; on slower remote
+        // (BrowserStack) devices this can exceed Espresso's default 26s idle timeout and
+        // throw IdlingResourceTimeoutException. Give the idle wait more headroom.
+        IdlingPolicies.setMasterPolicyTimeout(90, TimeUnit.SECONDS)
+        IdlingPolicies.setIdlingResourceTimeout(90, TimeUnit.SECONDS)
```
abolfazlimahdi, line 49: Nit: these raise the *global* Espresso idling policies and are never restored. Harmless under orchestrator-per-test isolation, but in a local non-orchestrator run they'd leak into every class that runs after this one in the same instrumentation invocation. Consider resetting them in `@After`, or leaving a comment that the leak is intentional.

bank-sdk/example-app/src/androidTest/scripts/bs_build_and_upload.sh
```diff
@@ -50,104 +70,140 @@ if [[ "$BS_USER" == "<your_browserstack_user_name>" ]] || [[ "$BS_KEY" == "<your
...
+  CLASS_JSON=""
+  for target in "${TEST_TARGETS[@]}"; do
+    case "$target" in
+      *.*) fq="$target" ;;               # already fully-qualified
+      *)   fq="$TEST_PACKAGE.$target" ;;  # short name -> prepend package
+    esac
+    if [ -n "$CLASS_JSON" ]; then CLASS_JSON="$CLASS_JSON, "; fi
+    CLASS_JSON="$CLASS_JSON\"$fq\""
+  done
+  FILTER_JSON="\"class\": [$CLASS_JSON],"
```
abolfazlimahdi, line 205: Nit: the trigger payload is built by string interpolation, so a malformed target argument (stray quote, space) produces broken JSON and a confusing API error. Fine for a dev script, but a cheap format check on each target (e.g. reject anything not matching `[A-Za-z0-9_.]+(#[A-Za-z0-9_]+)?`) would fail fast with a clear message.

abolfazlimahdi (commented):
## Review verdict: changes requested

_(GitHub does not allow a formal "Request changes" review on one's own pull request, so this is submitted as a comment review — please treat the two blocking issues below as changes requested before merge.)_

## Overview

This PR stabilizes the Bank SDK example-app UI suite for BrowserStack and adds sharded run tooling:

- **Test infra**: new `RetryRule` (retries transient network flakes), explicit `waitForExists` waits on network-dependent screens (extraction, digital invoice, no-result), raised Espresso idle timeouts for `OpenWithTest`, graceful handling of a missing `test.properties`, and `clickSkipButton` → `clickSkipButtonIfPresent` everywhere (correct, since onboarding only shows on first launch).
- **Test fixes**: `NoResultsTests` re-enabled with new no-result assets, BrowserStack-aware skip for the offline test, product-tag reset in `@After`.
- **Scripts**: `bs_build_and_upload.sh` now accepts multiple class/method targets, supports artifact reuse (build + upload once), and enables orchestrator flags; four shard scripts plus `bs_run_all_groups.sh`.

CI impact is contained: the example-app UI tests are commented out in `bank-sdk.check.ui-tests.yml`, so removing `@Ignore` from `NoResultsTests` only affects the BrowserStack scripts and local runs.

## What's good

- Comments consistently explain *why* (BrowserStack timing, flavor-dependent applicationId, skip-must-stay-skipped) — exactly what this kind of test infra needs.
- `RetryRule` correctly rethrows `AssumptionViolatedException` so skips aren't retried into failures.
- Orchestrator flags (`useOrchestrator`, `clearPackageData`, `disableAnimations`, `singleRunnerInvocation`) are the right call for flake reduction.
- `bs_run_all_groups.sh` is careful: temp file with `trap` cleanup, and parsing the URL file instead of `source`-ing it so its contents are never executed as shell code.
- The null-guards in the Espresso `check { view, _ -> ... }` callbacks are correct — Espresso passes a null view with a `NoMatchingViewException` when nothing matches.

## Blocking issues (see inline comments)

1. **`RetryRule` is not guaranteed to be the outermost rule** — `order = -1` equals `Rule.DEFAULT_ORDER`, so it ties with `activityScenarioRule` and the tie-break is effectively arbitrary for `@get:Rule` (method) rules. If the retry rule ends up inside the activity rule, retries re-run the test against the already-used activity, defeating the rule's purpose.
2. **`ErrorScreenTests`: only `test2` got the BrowserStack skip guard**, but `test3` and `test4` depend on the same network disconnect working. On BrowserStack they'll run online — burning 3 retry attempts before failing, or passing for the wrong reason.

## Non-blocking suggestions

- `DigitalInvoiceScreen` still hard-codes the applicationId in selectors while the PR fixes exactly that in `ExtractionScreen` / `DigitalInvoiceEditButton` — worth aligning via a small shared helper.
- `isInternetAvailable()` is checked ~2s after `svc wifi disable`; a short poll would make the skip decision reliable on slow local devices.
- The `runCatching { load test.properties }` + `as? String` block is now copy-pasted into four test classes — a shared helper would keep the "missing file means run" semantics in one place.
- Trailing newline still missing on `DigitalInvoiceEditButton.kt`, `DigitalInvoiceScreen.kt`, `NoResultScreen.kt` (it was fixed on `ExtractionScreen.kt`).
- The PR description is empty — given this changes how the whole suite runs (sharding, retries, skips), a short write-up of the sharding strategy and retry policy would help reviewers and future readers.

Solid, well-reasoned stabilization work overall — requesting changes for items 1 and 2 above.

bank-sdk/example-app/src/androidTest/java/net/gini/android/bank/sdk/exampleapp/ui/resources/RetryRule.kt
MozhganPeivandianSharbaf: Good catch — fixed. You're right that `Rule.DEFAULT_ORDER == -1` (confirmed in the junit-4.13.2 jar), so `order = -1` was effectively a no-op that just tied with `activityScenarioRule`. Changed it to `order = Int.MIN_VALUE` across all seven test classes (including `OpenWithTest`, for consistency), and updated the KDoc example + rationale to explain why.

bank-sdk/example-app/src/androidTest/java/net/gini/android/bank/sdk/exampleapp/ui/testcases/ErrorScreenTests.kt
MozhganPeivandianSharbaf: Fixed. Extracted a shared `skipIfNetworkCouldNotBeDisabled()` helper and applied it to `test2`, `test3`, and `test4` so none of them burn retries online on BrowserStack. Also addressed your pre-existing note in the same pass: `reconnectTheInternetConnection()` is now in an `@After`, so a failed or skipped test no longer leaves the device offline for the tests that follow.

bank-sdk/example-app/src/androidTest/java/net/gini/android/bank/sdk/exampleapp/ui/screens/ErrorScreen.kt
MozhganPeivandianSharbaf: Fixed. `isInternetAvailable()` now polls for up to ~5s: it returns `false` as soon as the network actually drops (the disconnect worked → don't skip), and only reports the network as still up if it stays up for the whole poll window. That removes the false-skip on slow local devices while still skipping correctly on BrowserStack.

bank-sdk/example-app/src/androidTest/java/net/gini/android/bank/sdk/exampleapp/ui/screens/DigitalInvoiceScreen.kt
MozhganPeivandianSharbaf: Fixed via a shared helper, as you suggested. Added `AppResources.resId(name)` (resolves the package from `targetContext.packageName`) and routed `DigitalInvoiceScreen` (all six selectors), plus `ExtractionScreen` and `DigitalInvoiceEditButton`, through it — so the flavor-safe resolution now lives in one place instead of being duplicated.

bank-sdk/example-app/src/androidTest/java/net/gini/android/bank/sdk/exampleapp/ui/testcases/OpenWithTest.kt
MozhganPeivandianSharbaf: Valid point. Leaving it as a small follow-up rather than growing this PR further: under the orchestrator's per-test isolation the raised idling policy doesn't leak, but you're right it would in a local non-orchestrator run. I'll either reset it in `@After` or add an explicit "intentional" comment. Happy to fold it into this PR if you'd prefer.

bank-sdk/example-app/src/androidTest/scripts/bs_build_and_upload.sh
MozhganPeivandianSharbaf: Fair — it's a dev-only script so I left input validation out for now, but the fail-fast check is cheap. I can add the `[A-Za-z0-9_.]+(#[A-Za-z0-9_]+)?` format guard on each target in this PR if you'd like it here rather than as a follow-up.

MozhganPeivandianSharbaf:
Thanks for the thorough review @abolfazlimahdi — really useful catch on the `RetryRule` ordering. Pushed fixes for both blocking issues and the main non-blocking ones:

- **[Blocking 1]** `RetryRule` → `order = Int.MIN_VALUE` in all 7 classes + KDoc (since `Rule.DEFAULT_ORDER == -1`).
- **[Blocking 2]** offline-skip guard now on `test2`/`test3`/`test4` via a shared helper; `reconnect` moved to `@After`.
- `isInternetAvailable()` now polls for the network to actually drop.
- New shared `AppResources.resId()` helper; `DigitalInvoiceScreen` / `ExtractionScreen` / `DigitalInvoiceEditButton` routed through it.

Deferred as optional follow-ups (noted inline): `OpenWithTest` idling-policy reset, the script target-format check, the `test.properties` loader helper, trailing newlines, and a PR description write-up. Happy to fold any of these into this PR if you'd like them here — otherwise I'll take them as follow-ups. Ready for another look.

PR #918, opened 2026-05-12
Pp 2311 improve qr code warning message

abolfazlimahdi:


MozhganPeivandianSharbaf (changes requested):
Thank you 👍 
looks good, could you check valid comments from copilet as well?

PR #924, opened 2026-05-19
Xpl 22 sonar qube capture sdk issues

abolfazlimahdi:


MozhganPeivandianSharbaf (approved):
Looks good, Thank you!

MozhganPeivandianSharbaf (approved):
Approved with two minor suggestions:
Add final to ReturnReasonsMapper to match the other utility mappers in the same package.
Remove the extra blank line before /* in the three mapper files.

abolfazlimahdi:
@MozhganPeivandianSharbaf Your first comment was valid and I have modified the ReturnReasonsMapper and made it final.
The second point about removing the first line of the Mappers, I think this is the Java/Kotlin convention of comments. If you still feel it is incorrect, please let me know which lines do you exactly refer to so I can investigate further. Thank you!

MozhganPeivandianSharbaf (approved):
Thank you Mahdi, Looks good to me!

PR #940, opened 2026-07-03
Pp 2736 add color customiztion flag in configuration screen

abolfazlimahdi:


abolfazlimahdi:
Done, not needed anymore

PR #944, opened 2026-07-10
fix(capture-sdk): Fix camera frame color on rotation

abolfazlimahdi:
PP-3219


MozhganPeivandianSharbaf (approved):
Looks good, Thank you

PR #945, opened 2026-07-10
fix(QRCodePopup): Allow button labels in unsupported QR dialog to wra…

abolfazlimahdi:
…p onto multiple lines

PP-3223


MozhganPeivandianSharbaf (approved):
Looks good, Thank you 👍

PR #947, opened 2026-07-13
Pp 3226 text on camera screen should be changed for camera screen after an option on this is not a payment qr code selected

MozhganPeivandianSharbaf:


capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragmentImpl.java
```diff
@@ -1093,6 +1098,12 @@ private boolean isOnlyQRCodeScanningEnabled() {
         return GiniCapture.getInstance().isOnlyQRCodeScanning() && GiniCapture.getInstance().isQRCodeScanningEnabled();
     }
 
+    private boolean isQRCodeScanningAvailable() {
+        return GiniCapture.hasInstance()
```
abolfazlimahdi: `isQRCodeScanningAvailable()` duplicates an existing helper
`FeatureConfiguration.isQRCodeScanningEnabled()` is already statically imported (line 135) and does exactly `hasInstance() && isQRCodeScanningEnabled()`. Simplify to:
```
private boolean isQRCodeScanningAvailable() {
    return isQRCodeScanningEnabled() && !mQRCodeScanningDisabledByUser;
}
```

abolfazlimahdi (changes requested):
Thank Mozhgan. Looks good in general. I have added one comment. The next thing is this one:

`onQRCodeReaderFail` state is not reflected in the new helper
When the QR detector fails to load (onQRCodeReaderFail(), line 297), `setQRDisabledTexts()` is called once — but `isQRCodeScanningAvailable()` still returns true, so any rotation or mode update reverts the titles/hint back to "invoice or QR" even though QR scanning is broken. Pre-existing, but the new helper is the natural place to fold in a mQRCodeReaderFailed flag;

abolfazlimahdi (approved):
Looks good. Thank you.

capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragmentImpl.java
MozhganPeivandianSharbaf: it is fixed, Thanks @abolfazlimahdi

PR #949, opened 2026-07-17
Pp 2363 improve android networking response model with optional session manager

abolfazlimahdi:
# PP-2363 — Optional SessionManager: authentication moved into the OkHttp layer + self-managed authentication

## Why

Clients who authenticate the Gini API requests themselves were still forced to implement a `SessionManager`, because the access token was hardcoded into every request via the `withAccessToken` / `accessToken`-taking methods. Their ask (payment provider integration):

> Are you able to move the usage of your own SessionManager from the "withAccessToken" function to your own httpclient?

This PR does exactly that: token handling moves out of the repositories/remote sources into the OkHttp layer, and consumers with their own authentication can opt out of the SDK's session handling entirely.

### 1. `GiniSessionInterceptor` — authentication in the OkHttp layer
- A new (internal) OkHttp application interceptor adds the `Authorization: Bearer …` header using the configured `SessionManager`. It is installed on the document/payment and tracking-analytics clients — **not** on the User Center client (which the session manager itself uses to fetch tokens).
- Requests which already carry an `Authorization` header pass through untouched (User Center basic auth, the deprecated `accessToken` overloads, consumer application interceptors).
- The session fetch is serialized by a mutex — the guarantee that parallel first-use uploads create only **one** anonymous Gini user is preserved.
- Session errors/cancellations and exceptions *thrown* by a consumer's `SessionManager` are mapped to `ApiException` / `SessionCancellationException` (both `IOException`s — OkHttp rethrows anything else on its dispatcher thread, which would crash the app) and surface as the same `Resource.Error` / `Resource.Cancelled` shapes as before.
- The User Center client gets its own OkHttp `Dispatcher`, so a consumer client running at its concurrent-request limit cannot deadlock the token fetch.

### 2. Self-managed authentication (the new opt-out)
- `setSelfManagedAuthentication(true)` on `GiniBankAPIBuilder`, `GiniHealthAPIBuilder` and `GiniCaptureDefaultNetworkService.Builder`: the SDK installs **no** authentication, requires **no** client credentials or `SessionManager`, creates **no** anonymous users. The consumer's `GiniHttpClientProvider` OkHttpClient adds the `Authorization` header itself (application or network interceptor — both work). The consumer's access token never passes through the SDK.
- Building without a custom `GiniHttpClientProvider` in this mode fails fast with an `IllegalStateException`; a leftover `SessionManager` or client credentials are ignored with a logged warning.

### 3. Backward compatibility — no breaking changes
- Every `accessToken`-taking method is kept as a `@Deprecated` overload with **wire-identical** behavior (it builds its own bearer header; the interceptor passes it through). `withAccessToken`, `accessTokenMutex` and `bearerHeaderMap` are deprecated but retained. Removal is scheduled for the next major version.
- The [binary-compatibility-validator](https://github.com/Kotlin/binary-compatibility-validator) plugin now gates all releasable modules: committed `.api` dumps + `apiCheck` in every check workflow, so any future public-API break fails CI. (See the new "Public API Compatibility" section in the README.)
- Wire-level tests (MockWebServer) pin the exact HTTP requests before/after the migration; behavior tests pin the `Resource` error shapes, the deprecated-overload contracts and the self-managed mode.

### 4. Observable behavior changes (documented in the integration guides)
- **API requests are no longer serialized.** Previously the access-token mutex was held for the *entire* request, so every document-manager call ran one at a time. Now only the session fetch is serialized and requests run genuinely in parallel.
- **`Resource.Error.exception` on session failures is now an `ApiException`** wrapping the original exception as its `cause` (message/status code/headers/body unchanged).

### 5. Housekeeping from the pre-merge review
- Extraction parsing consolidated into `ExtractionsParser` / `BankExtractionsParser` / `CompositeDocumentJson` (public mappers demoted to internal where no cross-module consumer needs them; unused `AnalysisResultMapper` removed).
- Builders hold only the application context (the interceptor closure keeps the builder alive — a passed Activity context must not leak).
- New-code coverage raised above the 80% quality gates (builder composition tests, repository method tests, interceptor tests incl. a dispatcher-deadlock regression test).
- `S1133` ("remove this deprecated code someday") is excluded from the Sonar gates, scoped to the files carrying the deliberate compatibility deprecations.

## CI notes for reviewers
- The Sonar scanner now requires Java 21 (fixed on main in #951; the `sonar` jobs run on JDK 21 — build/test jobs and the JVM 1.8 bytecode target are unchanged). If quality gates still show stale "new code" from unrelated files, the main-branch baselines need one `workflow_dispatch` re-analysis per module.
- `GiniCaptureFragmentTest.savePhotosLocally_shouldBeEnabled_whenConfigurationSetToTrue` (capture-sdk instrumented test) **fails identically on plain main** — pre-existing breakage, not a regression of this PR; it needs its own ticket.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bank-api-library/library/src/main/java/net/gini/android/bank/api/mapper/BankExtractionsParser.kt
```diff
@@ -0,0 +1,73 @@
...
+import org.json.JSONObject
+
+/**
+ * Internal use only. (Public visibility is required because the capture sdk consumes this
+ * parser from its own module.)
+ *
+ * Parses the Gini Bank API extractions response JSON into the Bank SDK's extraction models,
+ * including the Bank API specific return reasons.
+ */
+object BankExtractionsParser {
```
MozhganPeivandianSharbaf: `BankExtractionsParser` is a public `object` in a published module, and the KDoc justification ("the capture sdk consumes this parser") matches no file — the only caller is `BankApiDocumentRepository:37`, in this same module. That ships permanent Maven surface which can't be withdrawn later without a breaking change; mark it `internal` and re-run `apiDump`.

core-api-library/library/src/main/java/net/gini/android/core/api/mapper/ExtractionsParser.kt
```diff
@@ -0,0 +1,170 @@
...
+     * Parses a complete extractions response (the response body of
+     * `GET /documents/{id}/extractions`) into an [ExtractionsContainer].
+     *
+     * @param extractionsResponse the extractions response JSON
+     * @return The parsed [ExtractionsContainer] with specific and compound extractions.
+     * @throws JSONException If the JSON does not have the expected structure or contains invalid data.
+     */
+    @JvmStatic
+    @Throws(JSONException::class)
+    fun parseExtractionsContainer(extractionsResponse: JSONObject): ExtractionsContainer {
```
MozhganPeivandianSharbaf: `parseExtractionsContainer` has no caller outside `ExtractionsParserTest`, and the same is true of its `BankExtractionsParser` twin — only the other four functions are consumed across a module boundary, by `bank-api-library`. Drop it from both objects, and correct this KDoc to name `bank-api-library` rather than the capture sdk, which references neither parser.

core-api-library/library/src/main/java/net/gini/android/core/api/internal/GiniCoreAPIBuilder.kt
```diff
@@ -305,26 +353,45 @@ abstract class GiniCoreAPIBuilder<DM : DocumentManager<DR, E>, G : GiniCoreAPI<D
...
+     * When self-managed authentication is enabled (see [setSelfManagedAuthentication]) a
+     * [SessionManager] which always returns an error is used — even if a SessionManager was
+     * passed to the builder constructor: the SDK never requests sessions in that mode because
+     * the consumer's OkHttpClient authenticates the API requests.
      *
      * @return The SessionManager instance.
      */
     @Synchronized
     open fun getSessionManager(): SessionManager {
+        if (isSelfManagedAuthentication) {
```
MozhganPeivandianSharbaf, line 390: Optional: a `SessionManager` or client credentials passed to `GiniBankAPIBuilder` / `GiniHealthAPIBuilder` are discarded here with nothing logged — the promised warning exists only in `GiniCaptureDefaultNetworkService.Builder.warnAboutIgnoredAuthenticationConfig`. Moving that helper down into this builder would cover all three entry points.

core-api-library/library/build.gradle.kts
```diff
@@ -15,6 +16,18 @@ sonar {
...
+        // compatibility API deprecated in PP-2363: it is scheduled for removal with the
+        // next major version, so the reminder must not fail the quality gate until then.
+        property("sonar.issue.ignore.multicriteria", "deprecatedAccessTokenApi")
+        property(
+            "sonar.issue.ignore.multicriteria.deprecatedAccessTokenApi.ruleKey",
+            "kotlin:S1133"
+        )
+        property(
+            "sonar.issue.ignore.multicriteria.deprecatedAccessTokenApi.resourceKey",
+            "**/net/gini/android/core/api/Document*.kt"
```
MozhganPeivandianSharbaf: Optional: this pattern also silences S1133 for `DocumentManager.kt` and `DocumentService.kt` — the only other direct `Document*.kt` files in that package — plus any future one. Only `DocumentRepository.kt` and `DocumentRemoteSource.kt` carry these deprecations; a second multicriteria entry per file would match the tighter scope the bank and health modules use.

core-api-library/library/src/test/java/net/gini/android/core/api/DocumentRepositoryTest.kt
```diff
@@ -0,0 +1,691 @@
...
+        val repository = createRepository(sessionManager)
+
+        repository.getDocument("document-id-13")
+        repository.deleteDocument("document-id-13")
+
+        assertThat(sessionManager.getSessionCallCount.get()).isEqualTo(2)
+    }
+
+    @Test
+    fun `parallel api calls are serialized by the access token mutex`() = runTest {
```
MozhganPeivandianSharbaf: Optional: the assertions are right, but the name and the comment below both credit `accessTokenMutex`, which is deprecated and no longer on this path — the single-session guarantee comes from `GiniSessionInterceptor.sessionMutex`. Once `accessTokenMutex` goes in the next major this will read as a stale test.

```suggestion
    fun `parallel api calls are serialized by the session interceptor's mutex`() = runTest {
```

core-api-library/library/src/main/java/net/gini/android/core/api/http/GiniSessionInterceptor.kt
```diff
@@ -0,0 +1,107 @@
...
+        // session manager's cache, so the work is not wasted.
+        //
+        // TooGenericExceptionCaught: only IOExceptions may leave an interceptor - OkHttp's
+        // async path rethrows anything else on its dispatcher thread which crashes the app.
+        // A consumer's SessionManager should return Resource errors but may throw anything.
+        @Suppress("TooGenericExceptionCaught")
+        val sessionResource = try {
+            runBlocking {
+                sessionMutex.withLock {
+                    sessionManager.getSession()
```
MozhganPeivandianSharbaf, line 76: `getSession()` now runs once per HTTP request rather than once per repository operation, so a client's own `SessionManager` is called twice for `createPartialDocument` and 1+N times for `getPaymentProviders` — one per provider icon. `AnonymousSessionManager` caches internally so no test shows this; worth a third bullet in both guides' "Authentication behavior changes" sections.

core-api-library/library/src/test/java/net/gini/android/core/api/DocumentRemoteSourceWireTest.kt
```diff
@@ -0,0 +1,466 @@
+// This suite intentionally exercises the deprecated accessToken-taking methods: they must keep
+// setting the Authorization header themselves until they are removed (PP-2363).
+@file:Suppress("DEPRECATION")
```
MozhganPeivandianSharbaf, line 3: Every call here passes `accessToken`, so this suite pins only the deprecated overloads; of the 12 tokenless ones only `uploadDocument` is pinned elsewhere. A mismatch in a pair's `headerMap` arguments — `getFile`'s `accept = null`, `getPayment`'s positional contentType — would change the migrated request while these stay green. Bank and health wire tests use the new path; worth mirroring.

health-api-library/library/src/doc/source/guides/getting-started.rst
```diff
@@ -56,6 +56,39 @@ with this configuration would be ``550e8400-e29b-11d4-a716-446655440000@example.
...
+need client credentials or a ``SessionManager``. Enable self-managed authentication and pass your own
+``OkHttpClient`` which adds the ``Authorization`` header to the API requests (with an application or network
+interceptor - either works, because the library installs no authentication of its own in this mode). Your access
+token is never passed through the library and no anonymous Gini users are created:
+
+.. code-block:: java
+
+    val giniHealthApi: GiniHealthAPI =
+            GiniHealthAPIBuilder(context)
+                    .setSelfManagedAuthentication(true)
```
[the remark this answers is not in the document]
abolfazlimahdi: Fixed in 91b8a630b: the snippet is now declared as a kotlin code block (and the Java-style trailing semicolon is gone).

bank-api-library/library/src/doc/source/guides/getting-started.rst
```diff
@@ -77,6 +77,39 @@ with this configuration would be ``550e8400-e29b-11d4-a716-446655440000@example.
...
+need client credentials or a ``SessionManager``. Enable self-managed authentication and pass your own
+``OkHttpClient`` which adds the ``Authorization`` header to the API requests (with an application or network
+interceptor - either works, because the library installs no authentication of its own in this mode). Your access
+token is never passed through the library and no anonymous Gini users are created:
+
+.. code-block:: java
+
+    val giniBankApi: GiniBankAPI =
+            GiniBankAPIBuilder(context)
+                    .setSelfManagedAuthentication(true)
```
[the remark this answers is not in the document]
abolfazlimahdi: Fixed in 91b8a630b: the snippet is now declared as a kotlin code block (and the Java-style trailing semicolon is gone).

bank-api-library/library/src/main/java/net/gini/android/bank/api/mapper/BankExtractionsParser.kt
abolfazlimahdi: Fixed in 612fe34b0: BankExtractionsParser is internal and gone from the api dump. The stale KDoc justification was removed too — the capture sdk consumer it referred to (AnalysisResultMapper) had already been dropped in 05af07230, which orphaned this parser.

core-api-library/library/src/main/java/net/gini/android/core/api/mapper/ExtractionsParser.kt
abolfazlimahdi: Fixed in c60d25f04: parseExtractionsContainer is dropped from both objects and the KDoc now names the bank api library as the only cross-module consumer. Coverage is preserved — the tests now pin the remaining functions directly (parseCandidates/parseSpecificExtractions/parseCompoundExtractions in core, parseReturnReasons in bank).

core-api-library/library/src/main/java/net/gini/android/core/api/internal/GiniCoreAPIBuilder.kt
abolfazlimahdi: Fixed in e08c55b7b. Moved the warning into GiniCoreAPIBuilder.setSelfManagedAuthentication — since the SessionManager and client credentials can only arrive via the constructor, the ignored config is fully known at set-time, and this covers all three entry points. The capture builder now passes its authentication config through to the GiniBankAPIBuilder (which ignores it in self-managed mode), so its three-way construction collapsed into one call. One deliberate trade-off: the warning moved from capture's slf4j logger to android.util.Log (tag GiniCoreAPIBuilder), since core has no slf4j dependency and adding one for a log line didn't seem justified — it still lands in logcat, where this dev-time warning is read. The warnings are pinned by ShadowLog tests.

core-api-library/library/build.gradle.kts
abolfazlimahdi: Fixed in e5509d538: two multicriteria entries now target DocumentRepository.kt and DocumentRemoteSource.kt exactly — verified these are the only files in core's main sources carrying @Deprecated.

core-api-library/library/src/test/java/net/gini/android/core/api/DocumentRepositoryTest.kt
abolfazlimahdi: Fixed in 0d3e5d04d: renamed as suggested, and the comment inside the test now credits GiniSessionInterceptor.sessionMutex.

core-api-library/library/src/main/java/net/gini/android/core/api/http/GiniSessionInterceptor.kt
abolfazlimahdi: Fixed in a3a3a780e: both guides' "Authentication behavior changes" sections have a third bullet documenting the once-per-HTTP-request getSession() calls (the health guide also names getPaymentProviders' per-icon downloads) and advising a custom SessionManager to return a cached session while it is valid. Agreed on deferring interceptor-side caching until there's a 401-invalidation path.

core-api-library/library/src/test/java/net/gini/android/core/api/DocumentRemoteSourceWireTest.kt
abolfazlimahdi: Fixed in 83d3f6258: added DocumentRemoteSourceTokenlessWireTest mirroring the deprecated suite test-for-test with a stub auth interceptor (the bank/health pattern). It pins the headerMap arguments per overload — getFile's missing Accept, getPayment's positional contentType, getDocument's absent Content-Type — and also pins core's sendFeedback route via the annotated test service. 14 tests, all green without touching production code.

PR #950, opened 2026-07-17
HEAL-132 address Copilot issues on new target SDK for health

MozhganPeivandianSharbaf:
This pull request improves how window insets (such as system bars and keyboard) are handled in the payment review UI, ensuring correct padding and margin behavior across Android versions, especially when keyboard visibility changes or insets listeners are installed after initial view attachment. The changes address both the logic for applying insets and the reliability of insets updates.

**Insets handling improvements:**

* Updated the documentation and logic in `ReviewView` to clarify and improve handling of system insets, especially for Android 15+ where relying on raw IME insets could cause unwanted padding when the keyboard is visible. Now, keyboard visibility is observed manually, and for Android 14 and below, an attach-aware insets request is used to ensure padding is always applied, even if the insets listener is set after the initial dispatch.
* Added a new private function `requestApplyInsetsWhenAttached` in `ReviewView` to explicitly request an insets pass as soon as a view is attached to the window. This ensures that insets listeners installed after the initial dispatch still receive insets updates promptly.
* Applied `requestApplyInsetsWhenAttached` to relevant views (`gpsPaymentDetails`) to guarantee proper insets application.

**Margin and padding preservation:**

* In the `View.applyWindowInsetsWithTopPadding` extension, captured the initial margins of a view before applying insets, so that insets are added to existing margins rather than overwriting them. This prevents loss of intended spacing defined in XML or code.

**Minor code maintenance:**

* Added missing import for `android.graphics.Rect` to support margin calculations.

### DIFF
--- a/internal-payment-sdk/sdk/src/main/java/net/gini/android/internal/payment/review/reviewComponent/ReviewView.kt
+++ b/internal-payment-sdk/sdk/src/main/java/net/gini/android/internal/payment/review/reviewComponent/ReviewView.kt
@@ -174,10 +174,12 @@ class ReviewView(private val context: Context, attrs: AttributeSet?) :
      *
      * - No need to handle the bottom sheet case, system insets are applied automatically.
      * - For standard fragments:
-     *   - On Android 15+ (API 35+), applying insetter causes unwanted bottom padding when keyboard
-     *   is visible.
+     *   - On Android 15+ (API 35+), relying on the raw WindowInsetsCompat ime inset adds unwanted
+     *   bottom padding when the keyboard is visible.
      *     So we manually observe keyboard visibility and apply the correct height.
-     *   - On Android 14 and below, WindowInsetsCompat listener is used.
+     *   - On Android 14 and below, a WindowInsetsCompat listener is used, with an attach-aware
+     *   insets request so the padding is applied even when the listener is installed after the
+     *   initial insets dispatch.
      */
 
     private fun handleViewInsets() {
@@ -209,6 +211,7 @@ class ReviewView(private val context: Context, attrs: AttributeSet?) :
                     v.updatePadding(bottom = bottom)
                     insets
                 }
+                requestApplyInsetsWhenAttached(binding.gpsPaymentDetails)
             }
         }
     }
@@ -312,12 +315,23 @@ class ReviewView(private val context: Context, attrs: AttributeSet?) :
             insets
         }
 
+        requestApplyInsetsWhenAttached(view)
+    }
+
+    /**
+     * Triggers an insets pass for [target] as soon as this view is attached to the window.
+     *
+     * A freshly installed [ViewCompat.setOnApplyWindowInsetsListener] only runs on the next insets
+     * dispatch. When the listener is installed after the initial dispatch it may never run until an
+     * unrelated insets event occurs, so we explicitly request an insets pass once we are attached.
+     */
+    private fun requestApplyInsetsWhenAttached(target: View) {
         if (isAttachedToWindow) {
-            ViewCompat.requestApplyInsets(view)
+            ViewCompat.requestApplyInsets(target)
         } else {
             addOnAttachStateChangeListener(object : OnAttachStateChangeListener {
                 override fun onViewAttachedToWindow(v: View) {
-                    ViewCompat.requestApplyInsets(v)
+                    ViewCompat.requestApplyInsets(target)
                     v.removeOnAttachStateChangeListener(this)
                 }
 
--- a/internal-payment-sdk/sdk/src/main/java/net/gini/android/internal/payment/utils/extensions/View.kt
+++ b/internal-payment-sdk/sdk/src/main/java/net/gini/android/internal/payment/utils/extensions/View.kt
@@ -2,6 +2,7 @@ package net.gini.android.internal.payment.utils.extensions
 
 import android.app.Activity
 import android.content.Context
+import android.graphics.Rect
 import android.os.Build
 import android.util.TypedValue
 import android.view.KeyEvent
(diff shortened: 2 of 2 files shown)


internal-payment-sdk/sdk/src/main/java/net/gini/android/internal/payment/utils/extensions/View.kt
```diff
@@ -96,10 +97,20 @@ fun View.applyWindowInsetsWithTopPadding(
         if (navigationBars) typeMask = typeMask or WindowInsetsCompat.Type.navigationBars()
         if (statusBars) typeMask = typeMask or WindowInsetsCompat.Type.statusBars()
         val finalTypeMask = typeMask
+        // Capture the margins defined in XML/code once, so applying insets adds to them
+        // instead of overwriting them (which would collapse any intended spacing).
+        val initialMargins = (layoutParams as? ViewGroup.MarginLayoutParams)?.let {
```
abolfazlimahdi, line 102: `initialMargins` is snapshotted at call time, so if this extension is ever called a second time on the same view instance after insets were applied, the snapshot bakes in the previous insets and they double up. All current call sites call once per view lifecycle, so nothing is broken today — but storing the initial margins in a view tag (capture-if-absent) would make repeat calls idempotent. Fine as a follow-up.

internal-payment-sdk/sdk/src/main/java/net/gini/android/internal/payment/review/reviewComponent/ReviewView.kt
```diff
@@ -312,12 +315,23 @@ class ReviewView(private val context: Context, attrs: AttributeSet?) :
...
+    }
+
+    /**
+     * Triggers an insets pass for [target] as soon as this view is attached to the window.
+     *
+     * A freshly installed [ViewCompat.setOnApplyWindowInsetsListener] only runs on the next insets
+     * dispatch. When the listener is installed after the initial dispatch it may never run until an
+     * unrelated insets event occurs, so we explicitly request an insets pass once we are attached.
+     */
+    private fun requestApplyInsetsWhenAttached(target: View) {
```
abolfazlimahdi, line 328: Small nit: the attach-state listener is registered on the `ReviewView` (`this`) rather than on `target`. It works because both targets are children of this view, but `target.doOnAttach { ViewCompat.requestApplyInsets(it) }` from androidx-core-ktx does exactly this in one line, on the right view.

Also, since this is generic insets plumbing, consider moving it as an internal extension into `utils/extensions/View.kt` next to `applyWindowInsetsWithTopPadding` so other views in the module can reuse it.

internal-payment-sdk/sdk/src/main/java/net/gini/android/internal/payment/review/reviewComponent/ReviewView.kt
```diff
@@ -209,6 +211,7 @@ class ReviewView(private val context: Context, attrs: AttributeSet?) :
                     v.updatePadding(bottom = bottom)
                     insets
                 }
+                requestApplyInsetsWhenAttached(binding.gpsPaymentDetails)
```
abolfazlimahdi, line 214: With this change the bottom padding (`max(navBars, ime)`) is applied immediately at attach on API ≤ 34, instead of waiting for the first insets event. In non-edge-to-edge host activities the dispatched system-bar insets should be consumed to 0, so no visual change expected — but worth one manual check on an API 34 device in both modes since we have no test coverage here.

abolfazlimahdi (approved):
Looks good overall — the three Copilot findings on `internal-payment-sdk` are genuinely fixed (additive margins, attach-aware insets request on the pre-15 path, updated KDoc), and extracting `requestApplyInsetsWhenAttached` also fixes the old inline code that called `requestApplyInsets` on the wrong view. No public API changes, and since only health-sdk consumes internal-payment-sdk the blast radius is contained.

Two Copilot comments from #915 aren't covered by this diff — the missing `requestApplyInsets` in the example app's `ReviewActivity` and the `===` comparison in `DocumentPageAdapter`. Both look acceptable as-is to me, but since this PR's purpose is addressing the Copilot review, could you resolve those two threads on #915 with a short rationale?

One ask before merge: a quick on-device sanity check on API ≤ 34, since the bottom padding now applies immediately at attach instead of at the first insets event (both edge-to-edge and normal host activities). Also: typo in the PR title ("copilet"), and please confirm the internal-payment-sdk version bump is intentionally left to the release process.

internal-payment-sdk/sdk/src/main/java/net/gini/android/internal/payment/utils/extensions/View.kt
MozhganPeivandianSharbaf: Good catch, and agreed — the snapshot is taken at call time, so a second call on the same view instance after insets were already applied would capture margins that already include them and double up.

Confirmed it's latent only: all four call sites (`MainActivity`, `InvoicesActivity`, `OrdersActivity`, `OrderDetailsFragment`) call it once per view lifecycle on `binding.root`, and the function early-returns below API 35, so the exposure is 15+ only.

Agree the view-tag capture-if-absent is the right shape for idempotency. Taking it as a follow-up as you suggested rather than widening this PR.

internal-payment-sdk/sdk/src/main/java/net/gini/android/internal/payment/review/reviewComponent/ReviewView.kt
MozhganPeivandianSharbaf: Agreed on both.

On the first: the attach listener is registered on `this`, so `onViewAttachedToWindow(v)` receives the `ReviewView` and not `target`. It happens to work because `requestApplyInsets()` walks up the parent chain to the `ViewRootImpl` and re-dispatches to the whole tree, so the argument doesn't actually decide which view gets the pass — but you're right that it reads as if it does, which is worse than just being wrong. `doOnAttach` expresses the intent correctly and handles the already-attached case too. `core-ktx` is already on the classpath at 1.16.0, so it's a drop-in.

On the second: agreed, it's generic plumbing and belongs next to `applyWindowInsetsWithTopPadding` as an internal extension rather than living private in `ReviewView`.

Taking both as a follow-up rather than widening this PR — neither changes behaviour today, so I'd rather not add churn to a PR you've already approved. Shout if you'd prefer them in here.

internal-payment-sdk/sdk/src/main/java/net/gini/android/internal/payment/review/reviewComponent/ReviewView.kt
MozhganPeivandianSharbaf: Done — checked on an API 34 device in both modes.

Your reasoning matches what we saw: in the non-edge-to-edge host the system-bar insets arrive consumed, so there's no visual delta from applying them earlier. In the edge-to-edge host the bottom padding is correct at first render, and there's no jump from it landing at attach rather than at the first insets event. Keyboard open/close tracks as before. Spot-checked API 35 too — unchanged, as expected, since that path goes through the manual keyboard-observation branch.

Screenshots to follow. Agreed on the coverage gap being the reason this needed eyes rather than a test.

PR #954, opened 2026-07-27
feat(skill): Add skill to update Figma links in Confluence documentation

MozhganPeivandianSharbaf:
This pull request introduces a new skill, `update-figma-links`, for safely migrating Figma design links in the Android Gini Bank SDK Confluence documentation to a new UI-customisation guide. It provides a detailed, step-by-step process for updating, verifying, and standardizing Figma links, with strong guardrails to prevent errors. The PR also adds a supporting script, `transform_links.py`, which automates the link transformation and reformatting.

The most important changes are:

**New skill documentation and process:**
* [`.claude/skills/update-figma-links/SKILL.md`](diffhunk://#diff-933f08b0929313f07d4d4186bdb8d6f6b734bfe7ff388d74fefca66a5c05a0b1R1-R230): Adds comprehensive documentation for the `update-figma-links` skill, detailing when and how to use it, mandatory verification steps, guardrails, and handling of all Figma link formats in the Confluence page. The process ensures every migrated link points to the correct screen and includes human sign-off and post-write verification.

**Automation script for link migration:**
* `.claude/skills/update-figma-links/transform_links.py`: Introduces a Python script that automates the migration of Figma links in Confluence HTML. It supports:
  - Swapping old file keys and version numbers for new ones.
  - Stripping unnecessary share tokens from URLs.
  - Remapping node-ids for screens that moved.
  - Optionally standardizing all links to a uniform live-preview embed format.
  - Producing a detailed audit of all link changes for review.

These changes provide a robust, auditable, and user-friendly workflow for updating Figma links in documentation, minimizing the risk of broken or misdirected links.

### DIFF
--- a/.claude/skills/update-figma-links/SKILL.md
+++ b/.claude/skills/update-figma-links/SKILL.md
@@ -22,6 +22,23 @@ The user wants to point the Android Gini Bank SDK "Figma Links" documentation
 page at a newer Figma UI-customisation guide (new SDK version). The page contains
 many Figma links, one per screen/section.
 
+## How to run
+
+Type `/update-figma-links` and paste the new Figma guide URL, or just say in
+plain words: *"update the Android Figma links to this new guide: <URL>"*.
+
+You normally provide **only the new Figma guide URL** — the Confluence page is
+already known (see Repo defaults). Example:
+
+```
+/update-figma-links https://www.figma.com/design/<NEW_KEY>/Android-Gini-Bank-SDK-<version>-UI-Customisation
+```
+
+What happens next: it reads the page, verifies each screen's node-id in the new
+file, shows you a ✅/❌ table, and **pauses for your sign-off** (and asks you to
+paste the correct link for any screen that moved) before writing. You approve; it
+writes and then diffs to confirm. Nothing is written without your OK.
+
 ## Repo defaults (Android)
 
 This skill targets the **Android** Gini Bank SDK documentation. Use these unless
@@ -29,7 +46,9 @@ the user overrides them:
 
 - **Confluence page:** "Figma Links in public documentation" — page ID `435257345`
   (`https://gini.atlassian.net/wiki/spaces/GBSV/pages/435257345/Figma+Links+in+public+documentation`)
-- **Confluence site (cloudId):** `gini.atlassian.net`
+- **Confluence site:** `gini.atlassian.net` — pass this as the `cloudId` argument to
+  the Atlassian MCP tools (they accept a site hostname, so the literal cloudId UUID
+  is not needed here).
 - **Guide name pattern:** `Android-Gini-Bank-SDK-<version>-UI-Customisation`
 
 Because the page is already known, the user normally only needs to paste the
--- a/.claude/skills/update-figma-links/SKILL.md
+++ b/.claude/skills/update-figma-links/SKILL.md
@@ -1,0 +1,230 @@
+---
+name: update-figma-links
+description: >-
+  Update the Figma design links on the Android Gini Bank SDK "Figma Links in
+  public documentation" Confluence page to a new UI-customisation guide (new SDK
+  version). Verifies every node-id actually resolves to the CORRECT screen in the
+  new Figma file (not just that it exists), handles all three Confluence embed
+  formats, strips share tokens, and writes back safely with a post-write diff.
+  Use when migrating that Android page from one SDK version's Figma file to a
+  newer one — e.g. 4.0 → 4.4.
+---
+
+# Update Figma links on a Confluence documentation page
+
+Migrate all Figma links on a Confluence page from an old Figma file/version to a
+new one, correctly and safely. This encodes a process that was validated by hand;
+follow it in order and do not skip the verification or sign-off steps.
+
+## When to use
+
+The user wants to point the Android Gini Bank SDK "Figma Links" documentation
+page at a newer Figma UI-customisation guide (new SDK version). The page contains
+many Figma links, one per screen/section.
+
+## Repo defaults (Android)
+
+This skill targets the **Android** Gini Bank SDK documentation. Use these unless
+the user overrides them:
+
+- **Confluence page:** "Figma Links in public documentation" — page ID `435257345`
+  (`https://gini.atlassian.net/wiki/spaces/GBSV/pages/435257345/Figma+Links+in+public+documentation`)
+- **Confluence site (cloudId):** `gini.atlassian.net`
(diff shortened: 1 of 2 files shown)


.claude/skills/update-figma-links/SKILL.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Repo defaults (Android)
+
+This skill targets the **Android** Gini Bank SDK documentation. Use these unless
+the user overrides them:
+
+- **Confluence page:** "Figma Links in public documentation" — page ID `435257345`
+  (`https://gini.atlassian.net/wiki/spaces/GBSV/pages/435257345/Figma+Links+in+public+documentation`)
+- **Confluence site (cloudId):** `gini.atlassian.net`
+- **Guide name pattern:** `Android-Gini-Bank-SDK-<version>-UI-Customisation`
```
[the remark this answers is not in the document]
MozhganPeivandianSharbaf: it is fixed

.claude/skills/update-figma-links/transform_links.py
```diff
@@ -0,0 +1,165 @@
...
+
+Usage (re-point + standardise to embed previews):
+  python3 transform_links.py --in before.html --out after.html \
+      --new-key NEWKEY --old-version 4.0 --new-version 4.4 --strip-token --to-embed [--map ...]
+"""
+import argparse
+import re
+import sys
+
+FIGMA_URL_RE = re.compile(r'https://www\.figma\.com/[^\s"\'<>]+')
```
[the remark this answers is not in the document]
MozhganPeivandianSharbaf: it is fixed

abolfazlimahdi (commented):
Nice skill overall — the verify-before-write / sign-off / post-write-diff process is solid. I found three bugs in `transform_links.py` that can each produce a silently wrong link (which is exactly what the skill's guardrails are meant to prevent), so they're worth fixing before merge. Details inline.

.claude/skills/update-figma-links/transform_links.py
```diff
@@ -0,0 +1,168 @@
...
+def transform_value(url, args, node_map):
+    """Apply key/version/node-id/token changes to a single URL string."""
+    if args.old_key and args.new_key:
+        url = url.replace(args.old_key, args.new_key)
+    if args.old_version and args.new_version:
+        url = url.replace(
+            f'Android-Gini-Bank-SDK-{args.old_version}-UI-Customisation',
+            f'Android-Gini-Bank-SDK-{args.new_version}-UI-Customisation')
+    for old_node, new_node in node_map.items():
+        url = url.replace(f'node-id={old_node}', f'node-id={new_node}')
```
abolfazlimahdi: Two bugs in this `--map` replacement loop, both verified by running the code:

**1. Prefix collision.** Plain substring `replace` means a map entry whose old node-id is a prefix of another id on the page corrupts the longer one:

```
map 123-45=999-1  turns  node-id=123-456  into  node-id=999-16
```

Fix by anchoring the match, e.g.:

```python
url = re.sub(rf'node-id={re.escape(old_node)}(?=[&"\s<]|$)',
             f'node-id={new_node}', url)
```

**2. Chained replacements.** Because map entries are applied sequentially, `--map 1-2=3-4 --map 3-4=5-6` rewrites `node-id=1-2` → `3-4` → `5-6`. This is realistic when screens swap positions between versions. Fix with a single-pass `re.sub` using a callback that looks the id up in the dict once, instead of looping over entries.

.claude/skills/update-figma-links/transform_links.py
```diff
@@ -0,0 +1,168 @@
...
+    if args.old_key is None:
+        keys = re.findall(r'figma\.com/design/([A-Za-z0-9]+)', body)
+        if keys:
+            args.old_key = max(set(keys), key=keys.count)
+            print(f'[auto-detected old key] {args.old_key}')
+
+    all_urls_before = FIGMA_URL_RE.findall(body)
+    report = []
+
+    if args.to_embed:
```
abolfazlimahdi, line 140: **`--to-embed` mode silently skips Figma URLs outside the three wrapper formats.** In default re-point mode the `FIGMA_URL_RE.sub` pass covers the whole body, but in this branch re-pointing only happens inside the wrapper-conversion callbacks. A plain inline `<a href>` Figma link in a paragraph would keep the **old** file key, with no entry in the audit report to flag it. Since `--to-embed` is the documented default policy in SKILL.md, this is the mainline path.

Suggested fix: after the three wrapper passes, run the re-point `FIGMA_URL_RE.sub(repoint, body)` over the remaining body to catch stragglers (it's a no-op for already-converted URLs).

.claude/skills/update-figma-links/transform_links.py
MozhganPeivandianSharbaf: Good catches — both reproduced. Fixed in 9117fbdd8 by replacing the sequential substring loop with a single-pass anchored remap:

```python
if node_map:
    url = re.sub(r'node-id=([^&"\'\s<]+)',
                 lambda m: f'node-id={node_map.get(m.group(1), m.group(1))}', url)
```

The capture group matches a whole node-id (stops at `&`/`"`/`<`/space), so a mapped id can no longer bleed into a longer one (prefix collision), and each id is looked up in the map exactly once, so entries never chain (`1-2 → 3-4 → 5-6`). Verified: `--map 123-45=999-1` leaves `123-456` untouched, and `--map 1-2=3-4 --map 3-4=5-6` yields `3-4`.

.claude/skills/update-figma-links/transform_links.py
MozhganPeivandianSharbaf: Valid — a bare inline `<a href>` Figma link kept the old key and was missing from the audit. Fixed in 9117fbdd8 by adding a catch-all re-point pass after the three wrapper passes.

One important nuance on the suggested fix: I did **not** run an unguarded `FIGMA_URL_RE.sub(repoint, body)`, because re-running the remap over already-converted URLs can re-apply a map entry and re-introduce the chaining bug from the other thread. Instead the catch-all only touches URLs that still contain the **old** key, so already-converted wrapper URLs are skipped:

```python
def catch(mo):
    before = mo.group(0)
    if not args.old_key or args.old_key not in before:
        return before
    after = transform_value(before, args, node_map)
    if after != before:
        report.append((before, after, "inline"))
    return after
body = FIGMA_URL_RE.sub(catch, body)
```

Verified the inline straggler is now re-pointed and audited, and that a chained `--map` is not double-applied by this second pass.

abolfazlimahdi (approved):
Looks good, thanks Mozhgan

PR #955, opened 2026-07-28
fix(capture-sdk): Pin unsupported QR warning type only when first warning is shown

abolfazlimahdi:
## What

Fixes the unsupported-QR-code warning showing the wrong (stale) warning type on first launch by changing *when* the warning type is pinned for the session.

- `GiniCaptureViewModel` no longer pins `isUnsupportedQRCodeWarningEnabled` from the first DataStore emission — the configuration provider now always tracks the latest persisted value.
- The session pin is taken in a single place, lazily, when the camera screen shows the first unsupported-QR warning (`CameraFragmentExtension.isUnsupportedQRCodeWarningEnabled`).
- Updated the documentation comments on `QRCodePopup` and `UnsupportedQrWarningSessionPin` to describe the single pin site.
- Added `CameraFragmentExtensionTest` covering that the warning type stays constant within a session regardless of when the configuration arrives.

## Why

The first DataStore emission is the configuration cached by the *previous* session. Pinning the flag on that emission latched the stale value and ignored the fresh remote configuration fetched later in the current session — so every session ran with the previous session's warning type, and on first launch the new warning popup never showed even when remotely enabled.

Pinning only at the moment the first warning is actually shown keeps the in-session guarantee (the warning type cannot change mid-session) while letting each session use the freshest configuration available at that moment.

## Ticket

[PP-3241](https://ginis.atlassian.net/browse/PP-3241)

[PP-3241]: https://ginis.atlassian.net/browse/PP-3241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/capture-sdk/sdk/src/main/java/net/gini/android/capture/GiniCaptureViewModel.kt
+++ b/capture-sdk/sdk/src/main/java/net/gini/android/capture/GiniCaptureViewModel.kt
@@ -17,13 +17,13 @@ internal class GiniCaptureViewModel(
     init {
         // DataStore is the single source of truth for the configuration flags.
         //
-        // isUnsupportedQRCodeWarningEnabled must stay constant for the whole session. Normally
-        // this observer pins it to the first persisted value emitted by DataStore — not to a
-        // synchronous in-memory default: on a relaunch a persisted `true` may not have been
-        // loaded yet, and pinning to the cold default would incorrectly disable the warning for
-        // the entire run. The pin is first-wins and shared via UnsupportedQrWarningSessionPin:
-        // should an unsupported QR code be scanned before the first emission, the camera screen
-        // pins the session value instead (see CameraFragmentExtension).
+        // isUnsupportedQRCodeWarningEnabled must stay constant for the whole session, but it is
+        // deliberately NOT pinned here. The first DataStore emission is the value cached by the
+        // *previous* session; pinning it would latch that stale value and ignore the fresh remote
+        // configuration saved later in this session — every session would run with the previous
+        // session's flag. Instead the provider always tracks the latest persisted value, and the
+        // camera screen pins it at the moment the first unsupported-QR warning is actually shown
+        // (see CameraFragmentExtension.isUnsupportedQRCodeWarningEnabled).
         //
         // clientID and amplitudeApiKey are not persisted, so they are preserved from whatever the
         // provider already holds (empty on first launch, real after the API responds).
@@ -31,14 +31,10 @@ internal class GiniCaptureViewModel(
             clientConfigurationStorage.getConfiguration()
                 .filterNotNull()
                 .collect { config ->
-                    val pinnedQrCodeWarningEnabled = unsupportedQrWarningSessionPin.pinIfAbsent {
-                        config.isUnsupportedQRCodeWarningEnabled
-                    }
                     giniBankConfigurationProvider.update { current ->
                         config.copy(
                             clientID = current.clientID,
-                            amplitudeApiKey = current.amplitudeApiKey,
-                            isUnsupportedQRCodeWarningEnabled = pinnedQrCodeWarningEnabled
+                            amplitudeApiKey = current.amplitudeApiKey
                         )
                     }
                 }
--- a/capture-sdk/sdk/src/test/java/net/gini/android/capture/camera/CameraFragmentExtensionTest.kt
+++ b/capture-sdk/sdk/src/test/java/net/gini/android/capture/camera/CameraFragmentExtensionTest.kt
@@ -4,14 +4,17 @@ import com.google.common.truth.Truth.assertThat
 import net.gini.android.capture.di.getGiniCaptureKoin
 import net.gini.android.capture.internal.provider.GiniBankConfigurationProvider
 import net.gini.android.capture.internal.provider.UnsupportedQrWarningSessionPin
+import org.junit.After
 import org.junit.Before
 import org.junit.Test
+import org.koin.core.module.Module
 import org.koin.dsl.module
 
 class CameraFragmentExtensionTest {
 
     private lateinit var configurationProvider: GiniBankConfigurationProvider
     private lateinit var sessionPin: UnsupportedQrWarningSessionPin
+    private lateinit var koinTestModule: Module
     private lateinit var extension: CameraFragmentExtension
 
     @Before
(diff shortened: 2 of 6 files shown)


capture-sdk/sdk/src/test/java/net/gini/android/capture/camera/CameraFragmentExtensionTest.kt
```diff
@@ -0,0 +1,75 @@
...
+                    single { sessionPin }
+                }
+            )
+        )
+        extension = object : CameraFragmentExtension() {
+            override fun hideImageCorners() = Unit
+        }
+    }
+
+    @Test
```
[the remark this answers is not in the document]
abolfazlimahdi: Fixed in a822067f7: the test module is now held in a field and unloaded in @After, matching the cleanup pattern of the other tests in this module.

MozhganPeivandianSharbaf (commented):
Thanks for the quick turnaround on this — the fix itself looks right. Traced PP-3241's steps through the change: on the second launch the DataStore observer sets the provider to the cached `true`, the fresh remote configuration then re-emits through `GiniCaptureFragment.kt:152` and lowers it to `false`, and the first scan pins `false`, so the old yellow warning shows as the ticket expects. The inverse case (cached `false`, remote `true`) resolves correctly too, and the session boundary still holds — the ViewModel is `by viewModels()` on `GiniCaptureFragment`, so rotation keeps the pinned warning type and `onCleared()` releases it. No public API change, and nothing outside `capture-sdk` reads the flag, so the blast radius is contained.

Two things I would like addressed before merge, both inline: the code that actually fixes the bug (`GiniCaptureViewModel`) has no test — the three new tests pass unchanged if the pin is reinstated there, which is also why Sonar reports 0% coverage on new code — and the `@After` in `CameraFragmentExtensionTest` drops the production `UnsupportedQrWarningSessionPin` definition instead of restoring it, since Koin's `unloadModules` removes overridden definitions rather than reverting them.

One ask rather than a blocker: the pin now reads the provider at first-warning time, so it still depends on the network → DataStore → observer round-trip landing before the user scans. Practically that is seconds versus a fast request, but no unit test can cover it — could you run the ticket's steps once on a device on a real network? Worth deciding separately whether pinning the previous session's cached value is the intended fallback when the configuration request fails; happy for that to be a follow-up ticket.

capture-sdk/sdk/src/main/java/net/gini/android/capture/GiniCaptureViewModel.kt
```diff
@@ -17,28 +17,24 @@ internal class GiniCaptureViewModel(
...
         // clientID and amplitudeApiKey are not persisted, so they are preserved from whatever the
         // provider already holds (empty on first launch, real after the API responds).
         viewModelScope.launch {
             clientConfigurationStorage.getConfiguration()
                 .filterNotNull()
                 .collect { config ->
-                    val pinnedQrCodeWarningEnabled = unsupportedQrWarningSessionPin.pinIfAbsent {
-                        config.isUnsupportedQRCodeWarningEnabled
-                    }
                     giniBankConfigurationProvider.update { current ->
```
MozhganPeivandianSharbaf, line 34: Removing `pinIfAbsent` from this observer is the change that fixes PP-3241, and no test exercises it: all three tests in `CameraFragmentExtensionTest` drive `UnsupportedQrWarningSessionPin` and `GiniBankConfigurationProvider` directly, so they would pass unchanged if the pin were reinstated here and the bug could return unnoticed (Sonar reporting 0.0% coverage on new code is the same gap). Add a `GiniCaptureViewModelTest` that mocks `ClientConfigurationStorage.getConfiguration()` to emit the stale cached config (`isUnsupportedQRCodeWarningEnabled = true`) followed by the fresh one (`false`), then asserts `giniBankConfigurationProvider.provide().isUnsupportedQRCodeWarningEnabled` ends up `false`. `mockk` and `kotlinx-coroutines-test` — needed for `Dispatchers.setMain`, since `viewModelScope` dispatches on Main — are already on the capture-sdk unit-test classpath.

capture-sdk/sdk/src/test/java/net/gini/android/capture/camera/CameraFragmentExtensionTest.kt
```diff
@@ -0,0 +1,82 @@
...
+            override fun hideImageCorners() = Unit
+        }
+    }
+
+    @After
+    fun tearDown() {
+        // Restores the production singletons so the test instances don't leak into other test
+        // classes sharing the same Koin context.
+        getGiniCaptureKoin().unloadModules(listOf(koinTestModule))
+    }
```
MozhganPeivandianSharbaf, line 46: Koin's `unloadModules` does not restore an overridden definition — `InstanceRegistry.unloadModule` runs `_instances.remove(mapping)` for each index key the module declares — so this drops the production `single { UnsupportedQrWarningSessionPin() }` from `clientConfigurationModule` out of the process-wide isolated context instead of restoring it, which is the opposite of what the comment states. Nothing fails today because no other unit test in the module resolves that type, but the next one that does will fail with `NoDefinitionFoundException` depending on class execution order. Either re-declare the singleton after unloading, as below, or reload `clientConfigurationModule` itself — it is `internal` and visible from this source set once imported from `net.gini.android.capture.di`.

```suggestion
    @After
    fun tearDown() {
        // Koin's unloadModules drops the overriding definitions instead of restoring the
        // production ones, so the session pin is re-declared here: otherwise the shared Koin
        // context is left with no UnsupportedQrWarningSessionPin definition and every later
        // test in this JVM that resolves it fails with NoDefinitionFoundException.
        getGiniCaptureKoin().unloadModules(listOf(koinTestModule))
        getGiniCaptureKoin().loadModules(
            listOf(module { single { UnsupportedQrWarningSessionPin() } })
        )
    }
```

MozhganPeivandianSharbaf (approved):

capture-sdk/sdk/src/test/java/net/gini/android/capture/camera/CameraFragmentExtensionTest.kt
abolfazlimahdi: Resolved. Thanks for catching this.

PR #956, opened 2026-07-28
fix(capture-sdk): Fall back to old QR warning in QR-only mode

abolfazlimahdi:
## Summary

When the app is in **QR-code-scanning-only mode** (configured via `GiniCapture.Builder.setOnlyQRCodeScanning(true)`) and the user scans a non-payment QR code, the new 2-button warning dialog was shown. Its "Take photo of document" button is invalid in this mode because document capture is disabled. In QR-only mode the SDK now falls back to the old yellow warning popup.

The feature is gated behind the `unsupportedQRCodeWarningEnabled` client configuration flag, which is currently OFF in production — no user impact today, but this had to be fixed before the flag is enabled.

## Changes

- **`CameraFragmentExtension.kt`** — the session-pinned warning-type decision now also requires that QR-only mode is not enabled: `isUnsupportedQRCodeWarningEnabled && !isOnlyQRCodeScanningEnabled()`. The mode is part of the pin, so only one warning type appears per capture session: switching modes through the dialog's own buttons can only happen after the first warning was shown, i.e. after the decision was already pinned.
- **`CameraFragmentImpl.java`** — `isOnlyQRCodeScanningEnabled()` visibility changed from `private` to `protected` to implement the new abstract hook (logic untouched).

## Acceptance criteria

1. ✅ In QR-only mode, scanning an unsupported QR code no longer offers the invalid "Take photo of document" action.
2. ✅ In QR-only mode, the old yellow warning popup is shown instead.
3. ✅ Standard (non-QR-only) mode behavior is unchanged.

## Testing

- New unit tests in `CameraFragmentExtensionTest`: QR-only mode pins the old yellow warning even with the flag enabled; switching to QR-only via the dialog button keeps the new dialog for the rest of the session; the decision resets at the session boundary.
- `capture-sdk:sdk`: `testDebugUnitTest`, `lint`, `detekt`, `ktlintCheck` all pass locally.

PP-3286


capture-sdk/sdk/src/test/java/net/gini/android/capture/camera/CameraFragmentExtensionTest.kt
```diff
@@ -0,0 +1,120 @@
...
+        configurationProvider = GiniBankConfigurationProvider()
+        sessionPin = UnsupportedQrWarningSessionPin()
+        getGiniCaptureKoin().loadModules(
+            listOf(
+                module {
+                    single { configurationProvider }
+                    single { sessionPin }
+                }
+            )
+        )
```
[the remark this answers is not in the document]
abolfazlimahdi: This is resolved in another branch

capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragmentExtension.kt
```diff
@@ -45,13 +45,21 @@ internal abstract class CameraFragmentExtension {
...
+     * In QR-code-scanning-only mode the new dialog's "Take photo of document" action would be
+     * invalid (document capture is disabled), so the old yellow warning is pinned instead. The
+     * mode is part of the pinned decision: switching modes via the dialog's own buttons happens
+     * only after the first warning was shown, so it cannot change the warning type mid-session.
      */
     fun isUnsupportedQRCodeWarningEnabled(): Boolean =
         unsupportedQrWarningSessionPin.pinIfAbsent {
-            giniBankConfigurationProvider.provide().isUnsupportedQRCodeWarningEnabled
+            giniBankConfigurationProvider.provide().isUnsupportedQRCodeWarningEnabled &&
+                    !isOnlyQRCodeScanningEnabled()
```
MozhganPeivandianSharbaf, line 62: The pinned value now mixes the session-constant remote flag with `isOnlyQRCodeScanningEnabled()`, which reads mutable fragment state, so the mode observed at the first warning is frozen for the rest of the session. That is safe only while the two `mOnlyQRCodeScanningRuntimeOverride` writes stay reachable exclusively after the pin — an invariant nothing enforces, and one a future caller of `enableOnlyQRScanning()` would break silently. Pinning only the flag and evaluating the mode live (`pinIfAbsent { … } && !isOnlyQRCodeScanningEnabled()`) removes the coupling and still satisfies all three acceptance criteria, though it changes the behaviour after "Scan another QR code" — so it depends on the design answer in the summary.

MozhganPeivandianSharbaf (commented):
Looks good overall — the fallback does what PP-3286 asks. Traced `showUnsupportedQRCodePopup` → `isUnsupportedQRCodeWarningEnabled()` → `QRCodePopup.showViews()`: in QR-only mode with the flag on the pin resolves to `false` and the old yellow popup is shown, and standard mode still pins `true`, so AC3 holds. The `private` → `protected` widening adds nothing to the published surface, since `CameraFragmentExtension` is `internal` and `CameraFragmentImpl` is package-private — no version bump, and `bank-sdk` needs nothing.

One non-blocking point with no anchor line in the diff: `enableDocumentCapture()` (`CameraFragmentImpl.java:414`) sets `mOnlyQRCodeScanningRuntimeOverride = false` unconditionally, and that override wins over the integrator's `setOnlyQRCodeScanning(true)`. This PR closes the only path reaching it, but an early `return` when `isOnlyQRCodeScanning()` is true would make the invariant local. Follow-up is fine. The second point is inline.

Before merge — could you confirm on a QR-only build with the flag forced on that the yellow popup actually renders? The tests assert the decision, not the view. And one for design: after "Scan another QR code" in standard mode the pin keeps the new dialog, so "Take photo of document" is still offered. It works there, but it doesn't match AC1's wording and the linked Figma has no QR-only variant.

MozhganPeivandianSharbaf (approved):

capture-sdk/sdk/src/main/java/net/gini/android/capture/camera/CameraFragmentExtension.kt
abolfazlimahdi: Agreed that the pin now depends on the invariant that `mOnlyQRCodeScanningRuntimeOverride` is only written after the first warning was shown. I'd still keep the mode inside the pinned decision, for two reasons:

1. Evaluating the mode live would let the warning type flip mid-session (new dialog → old yellow popup) right after the user picks "Only QR scanning" in the dialog — which is exactly the inconsistency the session pin from PP-3241 exists to prevent.
2. After a *runtime* switch to QR-only, the dialog's "Take photo of document" action remains functional (it restores document capture, which the integrator allows). The action is only invalid in *integrator-configured* QR-only mode, and that case is already decided at the first warning, before any runtime switch can happen.

To make the invariant local instead of implicit, 9814c23bf adds the guard you suggested: `enableDocumentCapture()` now returns early when the integrator configured QR-only mode, so no future caller can silently override `setOnlyQRCodeScanning(true)`.

abolfazlimahdi:
Thanks for the thorough review! Responding to the points from the review summary:

**`enableDocumentCapture()` guard** — implemented in 9814c23bf rather than deferring to a follow-up: the method now returns early when the integrator configured QR-only mode (same condition as `isOnlyQRCodeScanningEnabled()`'s fallback path), so the runtime override can never re-enable document capture against `setOnlyQRCodeScanning(true)`. Added unit tests covering both the guard and the unchanged standard-mode switch-back.

**Manual verification** — good point that the tests only assert the decision, not the rendered view. I'll confirm the old yellow popup actually renders on a QR-only build with the flag forced on before merging.

**"Scan another QR code" in standard mode** — keeping the new dialog for the rest of the session is intended: one warning type per session (the PP-3241 invariant), and "Take photo of document" is valid and functional in standard mode. AC1 only constrains QR-only mode, which now falls back to the old popup — which is also why the missing QR-only variant in Figma doesn't block this PR: in QR-only mode we deliberately show the pre-existing old warning instead of a new design.

PR #959, opened 2026-07-31
chore: Add shared spec-feature skill with iOS sync workflows

abolfazlimahdi:
Split the spec-feature command into a platform-neutral SKILL.md and an Android-specific platform.md so the general workflow can be mirrored byte-identically in gini-mobile-ios without duplicating platform rules.

- .claude/skills/spec-feature/SKILL.md: general workflow (mirrored file, must stay identical to the iOS copy)
- .claude/skills/spec-feature/platform.md: Android conventions the skill loads at runtime (modules, architecture patterns, test stack, gates)
- shared-skills.check.yml: fails CI when mirrored files diverge from gini-mobile-ios (PRs, main pushes, weekly schedule); warns until the iOS counterpart exists
- shared-skills.sync.yml: on main pushes, opens/updates a sync PR in gini-mobile-ios (requires MOBILE_SKILLS_SYNC_TOKEN secret)
- mirrored-skills.txt: single list of mirrored files read by both workflows

### DIFF
--- a/.claude/skills/gini-build/SKILL.md
+++ b/.claude/skills/gini-build/SKILL.md
@@ -1,0 +1,81 @@
+---
+name: gini-build
+description: Implement a feature from its spec written by gini-spec-feature. Pass the ticket id as argument (e.g. PP-1234, IPC-42) — the spec must already exist in specs/.
+---
+
+<!--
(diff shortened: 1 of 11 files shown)


zladzeyka (approved):
LGTM!

PR #965, opened 2026-08-04
Pp 3262 redesign due date hint as bottom sheet

abolfazlimahdi:
## Pull Request Description

Redesigns the payment due date hint on the Analysis screen. The small inline tip card that auto-dismissed after 5 seconds is replaced by a modal warning bottom sheet (AlertDialog on tablets) — the same interaction pattern as the "document marked as paid" warning — giving the user an explicit choice: proceed with the transfer or cancel and return to the bank app. The inline card was easy to miss, offered no real choice, and conflicted with the rotating capture suggestions for TalkBack users.

PP-3262

**How:**

- `WarningType` gains a `PAYMENT_DUE_DATE` constant and now declares per-type primary and secondary CTA labels: for the due date hint "Proceed Anyway" is the primary CTA and "Cancel Transfer" the secondary one; the already-paid sheet keeps its existing hierarchy.
- `WarningBottomSheet` was generalized to be CTA-semantics-free: generic primary/secondary buttons labelled from the type, a listener with `onPrimaryAction`/`onSecondaryAction`, and a title format argument for the formatted due date. `AnalysisFragment` maps the CTAs to the actual behaviors per type (continue the flow vs. cancel the transaction) — the upcoming schedule-payment state only needs a label pair and a mapping case.
- The trigger logic in `AnalysisScreenPresenter.shouldShowPaymentDueHint()` is unchanged (client-config + SDK flag, threshold days, to-be-paid state, CX/RA/Skonto exclusions). Date formatting moved into an internal `DueDateFormatter` (`yyyy-MM-dd` → `dd.MM.yyyy`, raw-value fallback).
- While either warning sheet is visible, the rotating capture suggestions are stopped and their container hidden, so TalkBack focus stays on the sheet.
- The legacy Compose inline hint (`PaymentDueHintContent`, `PaymentDueHintColors`, `PaymentDueHintDismissListener`), its ComposeView containers in the three `gc_fragment_analysis.xml` variants and the obsolete strings are removed. These were public but implementation details of the hint; removal without a deprecation cycle was a ticket decision.

Sheet copy (EN + DE) comes from the "Scheduled Payment — Due Date Hint Redesign" Confluence page / Figma.

## Notes for Reviewers

- **Modules affected:** `capture-sdk:sdk` (all changes). `bank-sdk:sdk` has no code changes but consumes capture-sdk as a project dependency.
- **Verification:** local CI gate run — `testDebugUnitTest`, `lint`, `detekt`, `ktlintCheck` for `capture-sdk:sdk`, `capture-sdk:default-network` and `bank-sdk:sdk` — all green. Instrumented tests (`connectedCheck`) were not run locally; CI covers them.
- **Tests added/updated:**
  - `AnalysisScreenPresenterTest`: 13 new cases — shows the sheet at/beyond the threshold with the formatted date; not shown for below-threshold/today/past/missing/unparseable dates, disabled flags, wrong payment state, CX mode, Skonto extractions; already-paid precedence; proceed continuation returns extractions unchanged. Also adds a Koin test module (configuration provider + mocked `LastAnalyzedDocumentProvider`) — without it these code paths silently died on unresolved Koin definitions and earlier verifications passed vacuously.
  - `WarningBottomSheetTest` (new): per-type titles/descriptions/CTA labels, non-cancelable behavior, CTA clicks notify the listener and dismiss.
  - `DueDateFormatterTest` (new), `AnalysisFragmentTest` (+2: CTA-to-behavior mapping per type), `AnalysisFragmentImplTest` (+2: hints stopped and hidden for both sheets).
- **Test scenario:** with client config `paymentDueHintEnabled = true`, analyze an invoice whose `paymentDueDate` is ≥ 5 days in the future and `paymentState` is `ToBePaid` → the bottom sheet appears over the Analysis screen. "Proceed Anyway" (primary) continues to the extractions; "Cancel Transfer" clears stored images and returns to the host app. The sheet cannot be dismissed by tapping outside, dragging, or the back gesture. If the due date is today/past/below threshold, the flow proceeds directly.
- **Review per commit:** 5 incremental commits (spec → strings + formatter → bottom sheet feature → hints suppression → legacy removal); each commit builds and runs on its own.
- **Manual QA still open (per spec):** TalkBack focus order and announcement, dynamic font scaling, landscape + tablet dialog rendering, WCAG AA contrast, external keyboard navigation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


capture-sdk/sdk/src/main/java/net/gini/android/capture/analysis/AnalysisFragment.java
```diff
@@ -230,14 +230,15 @@ public void showAlertDialog(@NonNull final String message,
...
 
     @Override
-    public void showWarning(@NonNull WarningType type, @NonNull Runnable onProceed) {
+    public void showWarning(@NonNull WarningType type, @Nullable String titleFormatArg,
+                            @NonNull Runnable onProceed) {
         WarningBottomSheet sheet = (WarningBottomSheet)  fragmentManager.findFragmentByTag(WARNING_TAG);
         if (sheet == null) {
-            sheet = WarningBottomSheet.Companion.newInstance(type);
+            sheet = WarningBottomSheet.Companion.newInstance(type, titleFormatArg);
         }
```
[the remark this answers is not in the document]
abolfazlimahdi: Fixed in b55fcc518: the sheet found by tag is now reused only while it `isAdded()` (the state-restoration case); otherwise a fresh instance is created, so a stale instance can no longer show outdated arguments. Note this scenario was not reachable in practice — the presenter shows at most one warning type per analysis (else-if chain in `AnalysisScreenPresenter`) and both CTAs leave the screen — so this is defensive hardening.

capture-sdk/sdk/src/test/java/net/gini/android/capture/analysis/AnalysisScreenPresenterTest.kt
```diff
@@ -79,16 +89,45 @@ class AnalysisScreenPresenterTest {
...
+                    single { GiniBankConfigurationProvider() }
+                    single<LastAnalyzedDocumentProvider> {
+                        LastAnalyzedDocumentProvider(
+                            backgroundDispatcher = Dispatchers.IO,
+                            userAnalyticsEventTracker = get(),
+                        )
+                    }
+                }
+            )
+        )
```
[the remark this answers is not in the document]
abolfazlimahdi: Fixed in b55fcc518: `tearDown()` now reloads the SDK's real `providerModule` instead of duplicating its `LastAnalyzedDocumentProvider` definition. The `GiniBankConfigurationProvider` binding stays test-declared because it has no capture-sdk production module — it is registered by the Bank SDK's DI bridge.

capture-sdk/sdk/src/test/java/net/gini/android/capture/analysis/AnalysisScreenPresenterTest.kt
```diff
@@ -79,16 +89,45 @@ class AnalysisScreenPresenterTest {
...
+        // Bank SDK's DI bridge, so unit tests have to provide their own definition. The
+        // LastAnalyzedDocumentProvider is mocked because its production definition requires an
+        // initialized UserAnalytics tracker.
+        configurationProvider = GiniBankConfigurationProvider()
+        koinTestModule = module {
+            single { configurationProvider }
+            single<LastAnalyzedDocumentProvider> { mock() }
+        }
+        getGiniCaptureKoin().loadModules(listOf(koinTestModule))
     }
```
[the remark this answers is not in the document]
abolfazlimahdi: Not applicable: this project uses Koin 4.0.2, where `Koin.loadModules(modules, allowOverride = true, ...)` allows overriding definitions by default (since Koin 3.2). A `DefinitionOverrideException` is only thrown when `allowOverride = false` is passed explicitly, so the test setup cannot fail this way — the test suite runs green as-is.

capture-sdk/sdk/src/test/java/net/gini/android/capture/analysis/AnalysisScreenPresenterTest.kt
```diff
@@ -1134,4 +1170,314 @@ class AnalysisScreenPresenterTest {
...
+        // Then
+        val formattedDueDate = dueDate.format(DateTimeFormatter.ofPattern("dd.MM.yyyy"))
+        idleMainLooperUntil {
+            verify(mView).showPaymentDueHint(eq(formattedDueDate), any())
+        }
+        verify(listener, never()).onExtractionsAvailable(any(), any(), any())
+    }
+
+    @Test
+    fun `should show payment due hint sheet when remaining days equal the threshold`() {
```
MozhganPeivandianSharbaf, line 1193: **Blocker:** This asserts the sheet shows when `remainingDays == threshold`, matching `AnalysisScreenPresenter:628` (`< threshold`). The AC and the Confluence flow diagram both say `d <= threshold` shows no sheet, so the boundary is off by one against the spec — and this test now reads as the specification. Either tighten the presenter to `<=` and flip this case, or note that the deviation is deliberate.

capture-sdk/sdk/src/main/java/net/gini/android/capture/analysis/warning/WarningBottomSheet.kt
```diff
@@ -16,29 +16,33 @@ import net.gini.android.capture.R
...
- * Use [newInstance] with a [WarningType] to configure the title and description.
+ * Use [newInstance] with a [WarningType] to configure the title, description and CTA labels.
  */
 
 class WarningBottomSheet : BottomSheetDialogFragment() {
 
     interface Listener {
-        fun onCancelAction()
-        fun onProceedAction()
+        fun onPrimaryAction()
```
MozhganPeivandianSharbaf, line 39: **Blocker:** Renaming these on a public class breaks integrators: this branch has no `.api` dumps, no binary-compatibility-validator and no `explicitApi`, so `WarningBottomSheet` and `Listener` are published surface. Since the break is happening anyway, marking the class, `Listener` and `newInstance` `internal` here means it never has to happen again.

capture-sdk/sdk/src/main/java/net/gini/android/capture/analysis/warning/WarningType.java
```diff
@@ -7,22 +7,39 @@
...
  */
 public enum WarningType {
     DOCUMENT_MARKED_AS_PAID(
             R.string.gc_document_marked_paid_title,
-            R.string.gc_document_marked_paid_desc
+            R.string.gc_document_marked_paid_desc,
+            R.string.gc_cancel_transfer,
+            R.string.gc_proceed_anyway
+    ),
+    PAYMENT_DUE_DATE(
```
MozhganPeivandianSharbaf, line 28: **Suggestion:** `PAYMENT_DUE_DATE`'s title carries `%1$s`, but nothing on the type says the argument is required — `WarningBottomSheet.onCreate` only branches on `titleFormatArg != null`, so omitting it renders a literal `%1$s` with no crash or log. Not reachable today, but PP-3264's schedule state needs a date too. A `requiresTitleFormatArg` flag checked in `onCreate` would fail loudly instead.

capture-sdk/sdk/src/test/java/net/gini/android/capture/analysis/AnalysisScreenPresenterTest.kt
```diff
@@ -1134,4 +1170,314 @@ class AnalysisScreenPresenterTest {
...
+        }
+        verify(mView, never()).showPaymentDueHint(any(), any())
+    }
+
+    /**
+     * The presenter extension dispatches view calls through a background coroutine which
+     * posts back to the main looper, so the main looper is idled repeatedly until the
+     * verification passes (or the timeout is reached).
+     */
+    private fun idleMainLooperUntil(timeoutMs: Long = 5000, verification: () -> Unit) {
```
MozhganPeivandianSharbaf, line 1465: **Suggestion:** Better as a follow-up than widening this PR — this poll loop exists because `AnalysisScreenPresenterExtension:67` hardcodes `Dispatchers.IO`, so a genuine regression costs 5s per assertion across the 14 new cases. Injecting the dispatcher would let these use `kotlinx-coroutines-test` and drop the helper entirely.

capture-sdk/sdk/src/test/java/net/gini/android/capture/analysis/AnalysisScreenPresenterTest.kt
abolfazlimahdi: Decision made with product (see PP-1889 vs PP-3262 discussion): the behavior stays as implemented. PP-1889 — the original feature — specifies `paymentDueDate − today >= dueDateThreshold` shows the hint, so showing the sheet at exactly threshold days is the deliberate, pre-existing business logic (`AnalysisScreenPresenter` is unchanged since PP-1889, and PP-3262 itself states "business logic unchanged"). The "more than the threshold" / "≤ threshold → no sheet" wording in PP-3262's AC is the part that's off and will be aligned with PP-1889. No code change; this test correctly documents the intended boundary.

capture-sdk/sdk/src/main/java/net/gini/android/capture/analysis/warning/WarningBottomSheet.kt
abolfazlimahdi: Fixed in 9d379842e: `WarningBottomSheet` is now `internal` and additionally `@RestrictTo(LIBRARY)` (internal alone still compiles to public bytecode reachable from Java integrators; the annotation closes that with a `RestrictedApi` lint error). `WarningType` can't be `internal` (Java enum) nor package-private (referenced from `analysis` and `internal.ui`), so it got `@RestrictTo(LIBRARY)`. `Listener` and `newInstance` keep default visibility inside the internal class so the same-module Java callers aren't hit by internal-member name mangling. A usage sweep confirmed both types are referenced only within `capture-sdk:sdk`, and the PR description now has an "API changes" section consolidating this with the other surface changes (removed hint types, deprecated string keys) as input for the release notes.

abolfazlimahdi:
Closing out the two PR-level points from @MozhganPeivandianSharbaf's review:

- **PR title** — fixed: now `feat(capture-sdk): Redesign due date hint as a warning bottom sheet (PP-3262)`, matching the commit convention so the merge commit reads properly.
- **`specs/PP-3262-feature.md`** — decision: it stays committed. The `/gini-spec-feature` workflow writes specs there deliberately, and this one already earned its keep: it recorded the threshold-boundary semantics that the review discussion (PP-1889 vs PP-3262 AC) hinged on. We'll document the `specs/` convention in AGENTS.md as a follow-up outside this PR.

capture-sdk/sdk/src/test/java/net/gini/android/capture/analysis/AnalysisScreenPresenterTest.kt
abolfazlimahdi: Agreed — tracked as a follow-up in [PP-3406](https://ginis.atlassian.net/browse/PP-3406): inject the dispatcher into `AnalysisScreenPresenterExtension`, switch the presenter tests to `kotlinx-coroutines-test`, and drop the `idleMainLooperUntil` poll helper entirely.

capture-sdk/sdk/src/main/java/net/gini/android/capture/analysis/warning/WarningType.java
abolfazlimahdi: Implemented in c6fc03161 rather than deferred — it was small and scoped to files this PR already touches. `WarningType` now declares `requiresTitleFormatArg` (true for `PAYMENT_DUE_DATE`) and `WarningBottomSheet.newInstance` enforces it with `require()`, so a dated type created without its argument throws `IllegalArgumentException` at the call site instead of rendering a literal `%1$s`. Since the sheet is SDK-internal as of 9d379842e, the check can only ever hit SDK developers, never integrators. Covered by two new `WarningBottomSheetTest` cases.

